PR #16, opened 2022-05-31
Alert condition validation

giltk:
1. implemented the ratio alert type and the ratio block. 
2. fixes and cleaning to function alert resource read, create, and flatteners. 
3. updated the alert docs
4. updated the data source alert.


talknopf (approved):
Approved after code review in zoom

PR #30, opened 2022-09-04
fix: fix the goreleaser to start after semantic-release finish

giltk:


SirNatan1 (approved):
Reviewed the workflow which supposed to start when semantic-release finishes

PR #31, opened 2022-09-04
Rules update fix

giltk:


SirNatan1 (approved):
Overviewd the typos and added the use of /s in const.

PR #37, opened 2022-11-22
New terraform provider version (V1.3.0)

OrNovo:
New version of coralogix terraform provider


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
-@coralogix/team-cs-devops
\ No newline at end of file
+/.github/CODEOWNERS @coralogix/security-review
```
oded-dd, line 1: `/` should be removed as far as I know, otherwise not work. Also, we need to understand who are the people behind `security-review`

oded-dd (changes requested):

.github/CODEOWNERS
OrNovo: Your'e the only member of this group

oded-dd (dismissed):

oded-dd (approved):

PR #44, opened 2022-11-26
remove arithmeticOperatorModifier and remove 3H timeframe option

OrNovo:


oded-dd (approved):
I think you are missing 36H

PR #49, opened 2022-12-07
Documentation

OrNovo:
Updating documentation


oded-dd (changes requested):
Please also add that version 1.3 has major breaking changes, and customers that would like to upgrade it should expect changes. Link it with the issue opened.

oded-dd (approved):

PR #54, opened 2022-12-31
Coralogix dashboard

OrNovo:
Adding resource and data-source coralogix-dashboard


coralogix/resource_coralogix_dashboard.go
```diff
@@ -0,0 +1,1663 @@
...
+			Required:     true,
+			ValidateFunc: validation.StringIsNotEmpty,
+			Description:  "Dashboard name.",
+		},
+		"description": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "Dashboard description.",
+		},
+		"layout": {
```
oded-dd, line 1236: We need to be careful here, if the schema is breaking due to a change we did, we need to make sure we'll have a way to support backward compatible.

oded-dd: Also, don't we want to support file structure like we did in the enrichment (e.g. file = <layout.json>)

coralogix/resource_coralogix_dashboard_test.go
```diff
@@ -0,0 +1,284 @@
...
+				return fmt.Errorf("dashboard still exists: %s", rs.Primary.ID)
+			}
+		}
+	}
+
+	return nil
+}
+
+func testAccCoralogixResourceDashboard() string {
+	return fmt.Sprintf(`resource "coralogix_dashboard" test {
```
iravid, line 138: Wow this is really cool

examples/dashboard/main.tf
```diff
@@ -0,0 +1,157 @@
...
+              columns {
+                field           = "coralogix.metadata.subsystemName"
+                order_direction = "Unspecified"
+              }
+              columns {
+                field           = "coralogix.metadata.sdkId"
+                order_direction = "Unspecified"
+              }
+              columns {
+                field           = "textObject.textObject.textObject.log_obj.e2e_test.config"
```
iravid, line 138: @OrNovo i think the repeated textObject prefix can be dropped - could you check with the team?

coralogix/resource_coralogix_dashboard_test.go
OrNovo: Thanks 🙏

oded-dd (dismissed):

oded-dd (approved):

PR #102, opened 2023-05-19
patch -  change recoreding_rules_group to recoreding_rule_groups_set

OrNovo:


OrNovo:
@oded-dd I don't think it requires new minor version. WDYT?

PR #121, opened 2023-07-20
Update rest endpoints prefix

OrNovo:


oded-dd (approved):
I do think it should be through the ng-api endpoint but for now it should be enough

PR #256, opened 2024-08-19
Replace MS Teams webhook with MS Teams Workflow webhook

lootag:
### Description
Replace MS Teams webhook with MS Teams Workflow webhook


celaus (changes requested):
We need a statee upgrade for this or you roll back the rename

celaus (approved):

PR #264, opened 2024-08-23
Reintroduce old MS Teams webhook as readonly

lootag:
### Description
The MS Teams webhook was deprecated in 18.0.4, but it needs to stay readonly in order not to break compatibility with existing setups.


celaus:
Changelog

celaus:
Docs

celaus (approved):

PR #411, opened 2025-08-11
Use non-empty strings as resource IDs in order to avoid problems in Pulumi Terraform bridge.

### LINKED ISSUE #410: Static empty resource IDs causing problems in Pulumi
I understand this is not the most common scenario, but I am currently using Coralogix Terraform provider via [Pulumi Terraform Bridge](https://www.pulumi.com/registry/packages/terraform-provider/) (as per [Pulumi registry instructions](https://www.pulumi.com/registry/packages/coralogix/)).
Everything is working fine so far, but empty static `ID` fields used by three resources are causing problems.

### Affected Resource(s)

* coralogix_archive_logs
* coralogix_archive_metrics
* coralogix_archive_retentions

### Terraform Configuration Files

Please include all Terraform configurations required to reproduce the bug. Bug reports without a functional reproduction may be closed without investigation.

```hcl
<any usage of the above three resources>
```

Basically, any kind of usage of the three resources mentioned above generate the following error:
```
plugin for package 'terraform-provider' returned empty resource.ID from create <resource urn>.
```

I understand that this is not a problem of Coralogix provider in itself, and I did check things in [Pulumi repo](https://github.com/pulumi/pulumi).
I don't see empty IDs being supported in Pulumi any time soon, based on [this comment](https://github.com/pulumi/pulumi/issues/20186#issuecomment-3146480526):
> But the engine does (as I originally had suspected) treat an empty ID as a delete
> [...]
> Ideally we want to be consistent about using empty ID rather than empty state to indicate deletes (because there are valid cases where a refresh might not have any outputs to return but the resource does still exist). So I don't think there's actually anything to do here. **Just update the provider to return the ID when it should.**

The emphasis for last sentence is my own - essentially, the Pulumi maintainers are saying that providers must not return empty IDs as resource IDs.

Therefore, I think the path of "least resistance", sort to say, would be for Coralogix provider to use something else than empty strings as `ID` properties of those three resources.
Do you think that is feasible?
Is there any documentation that empty resource IDs would be dangerous also in Terraform itself?

### Expected Behavior

The three resources mentioned should've been successfully created.

### Actual Behavior

The three resources returned empty `ID` field, and resulted in crash.

### Steps to Reproduce

This needs Pulumi to reproduce - but I think it is also pretty clear what is the root problem here.
The code visible in [this search](https://github.com/search?q=repo%3Acoralogix%2Fterraform-provider-coralogix%20%22types.StringValue(%5C%22%5C%22)%22&type=code) will need modification to use something else than empty string as ID - perhaps static string, or maybe account name, or something else?

### Community Note

stazz:
### Description

This fixes #410 .

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?
- [ ] Have you run the acceptance tests on this branch?

I haven't run acceptance tests, because the `CONTRIBUTING.md` says:
```
Acceptance tests create real resources and in a case of failure some resources may not be removed.
```

And as our company's Coralogix settings already are controlled by this provider, I don't think I could run the acceptance tests against that.
Also the changes I made are pretty simple, and don't affect any logic utilizing invoking Coralogix SDK client.

But let me know if I need to create some freebie personal throwaway account for Coralogix in order to run the tests - I am fine with that too! 👍 

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

### Release Note

```release-note
Changed the Terraform resource ID from empty static string to non-empty static string for `coralogix_archive_logs`, `coralogix_archive_metrics`, and `coralogix_archive_retentions` resources.
This was done in order to accomodate usecase where this provider is used in Pulumi - there, empty resource ID signals resource deletion, and creating new resources with empty resource ID is not possible.

The change might cause extra noise in diff on first `plan` or `apply` run after update, but otherwise should not have any effect.
```

### Community Note

### DIFF
--- a/coralogix/resource_coralogix_archive_logs.go
+++ b/coralogix/resource_coralogix_archive_logs.go
@@ -152,7 +152,7 @@ func (r *ArchiveLogsResource) Create(ctx context.Context, req resource.CreateReq
 	}
 	log.Printf("[INFO] Submitted new archive-logs: %s", protojson.Format(createResp))
 
-	plan = flattenArchiveLogs(createResp.GetTarget(), RESOURCE_ID)
+	plan = flattenArchiveLogs(createResp.GetTarget(), RESOURCE_ID_ARCHIVE_LOGS)
 	if diags.HasError() {
 		resp.Diagnostics.Append(diags...)
 		return
@@ -279,4 +279,4 @@ func (r *ArchiveLogsResource) Delete(ctx context.Context, req resource.DeleteReq
 }
 
 // Safeguard against empty ID string, as using empty string causes problems when this provider is used in Pulumi via https://github.com/pulumi/pulumi-terraform-provider
-const RESOURCE_ID string = "archive-logs-settings"
+const RESOURCE_ID_ARCHIVE_LOGS string = "archive-logs-settings"
--- a/coralogix/resource_coralogix_archive_metrics.go
+++ b/coralogix/resource_coralogix_archive_metrics.go
@@ -218,7 +218,7 @@ func (r *ArchiveMetricsResource) Create(ctx context.Context, req resource.Create
 		return
 	}
 	log.Printf("[INFO] Received archiveMetrics: %s", protojson.Format(readResp))
-	plan, diags = flattenArchiveMetrics(ctx, readResp.GetTenantConfig(), RESOURCE_ID)
+	plan, diags = flattenArchiveMetrics(ctx, readResp.GetTenantConfig(), RESOURCE_ID_ARCHIVE_METRICS)
 	if diags.HasError() {
 		resp.Diagnostics.Append(diags...)
 		return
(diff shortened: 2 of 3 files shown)


coralogix/resource_coralogix_archive_logs.go
```diff
@@ -172,7 +172,8 @@ func flattenArchiveLogs(target *cxsdk.Target) *ArchiveLogsResourceModel {
 	}
 
 	return &ArchiveLogsResourceModel{
-		ID:                types.StringValue(""),
+		// Use non-empty ID string, as using empty string causes problems when this provider is used in Pulumi via https://github.com/pulumi/pulumi-terraform-provider
+		ID:                types.StringValue("archive-logs-settings"),
```
rshade, line 176: @stazz the id is already computed in 206, I would pass it to the function, and then add it to the ID here: https://github.com/coralogix/terraform-provider-coralogix/blob/7cfbee54b84717703d2b7b6f05fc6183390b8ab7/coralogix/resource_coralogix_archive_logs.go#L206C1-L206C30

stazz:
Hmm, looks like some silly errors due to my changes... Let me fix those, and I guess I could try running `make testacc` without any tokens locally to at least make it build the code successfully 👍

coralogix/resource_coralogix_archive_logs.go
stazz: Based on recent comment from @celaus , I will resolve this 👍

celaus (approved):

PR #480, opened 2026-03-04
Cx 27438 fix name error in rr

NoyaItzhaki:
### Description

Fixes recording rules group set handling so the Terraform `name` attribute is used correctly on create/update and name changes are applied in place.

* **Create/update from YAML:** When `name` is set in config, the provider sends it to the API instead of the top-level name from the YAML, avoiding `duplicate_rule_group_set_name` when the YAML uses a fixed name (e.g. "Example").
* **Update name in place:** The `name` attribute no longer has `RequiresReplace()`, so changing only `name` triggers an Update (same resource) instead of destroy + create.
* **Update from YAML:** `expandUpdateRecordingRulesGroupsSet` passes `plan.Name` into `expandRecordingRulesGroupsSetFromYaml`, so updates use the Terraform `name`; the redundant `updateName` conditional was removed per review.
* **CheckDestroy:** Added a nil check for `testAccProvider.Meta()` before the type assertion to avoid a panic when the destroy check runs after teardown.

**Test (`TestAccCoralogixRecordingRulesGroupsSetUpdateName`):** The test verifies that changing only `name` results in an in-place update (same resource ID). Step 1's first check iterates over Terraform state, finds the `coralogix_recording_rules_groups_set` resource by type, and stores its primary ID in `idAfterCreate` (lookup by type keeps the test valid regardless of state key shape, e.g. with ProtoV6). Step 2's first check is a closure that runs when Step 2's checks execute; it finds the same resource in state and asserts that its `id` equals `idAfterCreate`. A closure is used instead of `resource.TestCheckResourceAttr(..., "id", idAfterCreate)` because the test case is built once at startup when `idAfterCreate` is still empty; the closure runs after Step 1 has set `idAfterCreate`, so the comparison is correct. Together these blocks confirm the name change was applied via Update (same ID), not replace.

### Acceptance tests

```
$ make testacc TESTARGS='-run=TestAccCoralogixRecordingRulesGroupsSet'

...
```

### Release Note

```release-note
coralogix_recording_rules_groups_set: name attribute can be updated in place; Terraform name is used on create/update when provided (fixes duplicate_rule_group_set_name when using YAML with a fixed name).
```

### Community Note


internal/provider/recording_rules/resource_coralogix_recording_rules_groups_set.go
```diff
@@ -675,14 +675,18 @@ func expandRecordingRulesGroupsSet(ctx context.Context, plan *RecordingRuleGroup
...
+		setName := plan.Name.ValueString()
+		rrg, diags := expandRecordingRulesGroupsSetFromYaml(yamlContent, setName)
 		if diags.HasError() {
 			return nil, diags
 		}
-
+		updateName := rrg.Name
+		if len(setName) > 0 {
+			updateName = &setName
+		}
```
assafad1: Aren't `rrg.Name` and `setName` identical at this point? See `expandRecordingRulesGroupsSetFromYaml` func.

assafad1: So I think this part isn't needed. Correct me if I'm missing something.

internal/provider/resource_coralogix_recording_rules_groups_set_test.go
```diff
@@ -115,10 +120,23 @@ func TestAccCoralogixRecordingRulesGroupsSetUpdateName(t *testing.T) {
...
+								continue
+							}
+							if rs.Primary == nil || rs.Primary.ID == "" {
+								return fmt.Errorf("resource %s has no primary id", resName)
+							}
+							idAfterCreate = rs.Primary.ID
+							return nil
+						}
+						return fmt.Errorf("no coralogix_recording_rules_groups_set resource in state")
+					},
```
assafad1, line 137: could you please add to the PR description what these parts are doing?

assafad1 (approved):

PR #498, opened 2026-04-06
[CDS-2851] Fix OpenAPI domain base URL and add CX510, CX520 regions

daidokoro:
Fixes an issue where custom domains without the api. prefix would fail to connect to OpenAPI endpoints (such as TCO policies and Scopes). Also adds support for the CX510 and CX520 regions.


internal/provider/provider.go
```diff
@@ -89,6 +91,8 @@ var (
 		"US1":     "US1",
 		"USA2":    "US2",
 		"US2":     "US2",
+		"CX510":   "cx510.coralogix.com",
+		"CX520":   "cx520.coralogix.com",
```
oded-dd, line 95: BYOC - (cx510, cx520) - should not be defined alias but custom one.

daidokoro:
PR Created by AI in error. Closing.

PR #510, opened 2026-04-27
fix(BUGV2-5323): coralogix_alert not deleting webhooks_settings on update

royfur:
## Summary

  Fixes [BUGV2-5323](https://coralogix.atlassian.net/browse/BUGV2-5323): when a `coralogix_alert` resource transitions from `webhooks_settings` to `router` (or to an empty `notification_group`), the old webhooks were silently preserved
  instead of being deleted.

  > **Update after review:** an earlier version of this PR also included an API-layer change in `extractWebhooksSettings` (returning `[]` instead of `nil` for null/unknown inputs). Empirical validation against EU2 (see comment thread)
  confirmed the alerts API treats absent and `[]` as equivalent for proto3 repeated fields with no field mask, so that layer was unnecessary. Reverted in follow-up commits. The PR's effective diff is now just the schema-layer (Issue 1)
  fix.

  ---

  ## Root Cause

  ### Issue 1 — Terraform plan layer: `Computed: true` + `UseStateForUnknown()` on `webhooks_settings` (`alert_schema/v3.go`)

  In Terraform Plugin Framework, an `Optional + Computed` attribute with a null config value is planned as **unknown** (provider-computed). The `setplanmodifier.UseStateForUnknown()` modifier then replaces that unknown with the **prior
  state value** (the existing webhooks). The net effect: when a user removes `webhooks_settings` from their config, the plan silently keeps the old webhooks — no diff is shown, and the update call re-sends them to the API.

  This was introduced in commit `7b37189` ("NC and other Updates", Nov 2025) as part of the Notification Center feature, likely to avoid disrupting existing webhook config when NC fields were added alongside. The v2 schema never had
  this problem — `webhooks_settings` was `Optional` only.

  ### ~~Issue 2 — API layer: `nil` vs empty slice in `extractWebhooksSettings` (`resource_coralogix_alert.go`)~~

  ~~Even after fixing the plan layer, when `webhooks_settings` is null/removed the function returned Go `nil`. The SDK's `AlertDefNotificationGroup.ToMap()` uses `IsNil()` which returns `true` for a nil slice, causing the `webhooks` key
   to be **omitted entirely** from the JSON body. In a `ReplaceAlertDef` (full replace) call this is risky: an absent field may be treated as "no change" by some API implementations. Returning an empty non-nil slice
  `[]AlertDefWebhooksSettings{}` causes `IsNil()` to return `false`, so `"webhooks": []` is explicitly sent — an unambiguous signal to clear them.~~

  > _Initially identified and addressed; empirical testing on EU2 showed this layer was unnecessary — the alerts API treats absent and `[]` as equivalent for proto3 repeated fields with no field mask. Reverted in follow-up commits._

  ---

  ## Why `Computed: true` + `UseStateForUnknown` was added

  Added in `7b37189` ("NC and other Updates") when `router` and `destinations` (Notification Center) were introduced. The intent was to prevent the plan from showing a spurious diff on `webhooks_settings` when users started adding NC
  fields alongside existing webhooks. It achieved that but created the inverse bug: webhooks can never be removed via Terraform.

  ---

  ## Is this a breaking change?

  **No** for users with `webhooks_settings` explicitly in their config — behaviour is unchanged.

  **Corrected** for the reported scenario: webhook → router/empty now works as expected.

  **One edge case to be aware of**: after `terraform import`, if the API has webhooks the user hasn't written into config, the next `terraform plan` will show them as pending deletion. This is *correct* behaviour for an `Optional`
  (non-`Computed`) attribute and matches the v2 schema behaviour.

  Sub-fields `retriggering_period` and `notify_on` retain their own `Computed: true` + `UseStateForUnknown()` plan modifiers (and master's relaxation of `notify_on` to `Optional+Computed` is preserved through the merge), so computed
  defaults within a webhook entry still work correctly and existing configs continue to work without change.

  ---

  ## Risks

  - **Perpetual diff**: removed by the fix — sub-field modifiers handle this.
  - **API not clearing on absent field**: empirically verified non-issue — the alerts API treats absent and `[]` as equivalent for this proto3 repeated field, so `nil` (omitted JSON key) correctly clears webhooks server-side. The
  Issue-2 layer was reverted.
  - **Existing users with webhooks in config**: no change in behaviour.
  - **Schema version bump**: not required — removing `Computed` from an `Optional+Computed` attribute is non-breaking for Terraform state (the stored value type is unchanged).

  ---

  ## Changes

  | File | Change |
  |------|--------|
  | `internal/provider/alerts/alert_schema/v3.go` | Remove `Computed: true` and `setplanmodifier.UseStateForUnknown()` from `webhooks_settings`; remove now-unused `setplanmodifier` import |
  | `internal/provider/resource_coralogix_alert_test.go` | Add `TestAccCoralogixResourceAlert_webhooks_settings_deletion` covering the 3-step lifecycle: webhooks → router → no notifications; fix test fixture to include required
  `retriggering_period` alongside `notify_on` in `webhooks_settings` (API requires both or neither) |
  | `internal/provider/resource_coralogix_parsing_rules_test.go` | Add `ImportStateVerifyIgnore: []string{"order"}` to `TestAccCoralogixResourceParsingRules_parse` — the group `order` is server-assigned and can differ between create and
   import in a shared test environment |

  ## Test plan

  - [ ] `TestAccCoralogixResourceAlert_webhooks_settings_deletion` — new test, 3 steps: create with webhooks, update to router only (assert webhooks cleared), update to no notification group (assert webhooks still 0)
  - [ ] `TestAccCoralogixResourceAlert_logs_less_than_with_routing` — existing test now asserts `notification_group.webhooks_settings.#` = 0 after transitioning to router-only config
  - [ ] `TestAccCoralogixResourceParsingRules_parse` — import verify now ignores `order` (server-assigned, not user-controlled)
  - [ ] All existing alert acceptance tests pass

[BUGV2-5323]: https://coralogix.atlassian.net/browse/BUGV2-5323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmenilo1:
@royfur — addressing the Codex P1 above. 
Pushed a follow-up commit that splits IsNull() from IsUnknown() in extractWebhooksSettings: unknown returns nil (omit the JSON key, preserve server value until the reference resolves), null returns []AlertDefWebhooksSettings{} (explicit clear).
Validated locally end-to-end against EU2 — webhook → router → empty notification_group all transition cleanly, idempotency check passes after each step. Original BUGV2-5323 fix still works.

 Also opened a thought on follow-up: the same Optional + Computed + UseStateForUnknown pattern is still on group_by_keys, destinations, and labels in alert_schema/v3.go. 
Probably worth a separate PR rather than expanding scope here.

dmenilo1 (approved):

assafad1:
@codex review

internal/provider/alerts/resource_coralogix_alert.go
```diff
@@ -363,9 +362,12 @@ func extractNotificationGroup(ctx context.Context, notificationGroupObject types
 }
 
 func extractWebhooksSettings(ctx context.Context, webhooksSettings types.Set) ([]alerts.AlertDefWebhooksSettings, diag.Diagnostics) {
-	if webhooksSettings.IsNull() || webhooksSettings.IsUnknown() {
+	if webhooksSettings.IsUnknown() {
 		return nil, nil
 	}
+	if webhooksSettings.IsNull() {
+		return []alerts.AlertDefWebhooksSettings{}, nil
+	}
```
assafad1, line 368: AFAIK the alerts API removes configs that aren't included in the replace request. Webhooks is a proto3 repeated field with no field mask, so an absent field and [] should be indistinguishable to the server. If that's right, the schema fix alone should be sufficient and this change is a no-op. 
Did you verify it's actually needed (e.g., test the schema fix without this one)?

dmenilo1: Reverted d557e4b after empirically testing your hypothesis — schema fix alone correctly clears the webhook (no-op confirmed). 
Left the revert as a separate commit rather than rewriting history so the experiment is visible in the trail.

PR #512, opened 2026-04-28
 fix(BUGV2-5395): prevent panic in coralogix_archive_retentions when retentions come from a variable

dmenilo1:
### Description

  Fixes [BUGV2-5395](https://coralogix.atlassian.net/browse/BUGV2-5395): `terraform plan/apply` panics with `runtime error: index out of range [0] with length 0` when `coralogix_archive_retentions.retentions` is supplied through a
  Terraform variable. Inline HCL literals were unaffected.

  ## Root Cause

  Two compounding bugs in `retentionsValidator.ValidateList` (`internal/provider/dataengine/resource_coralogix_archive_retentions.go`):

  ### Bug 1 — null/unknown branch reports an error but doesn't `return`

  ```go
  if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
      resp.Diagnostics.AddError("error on validating retentions", "retentions can not be null or unknown")
  }   // ← falls through
  ```

  When `retentions = var.retentions_list`, Terraform calls `ValidateResourceConfig` with an unknown `ConfigValue` during the plan phase. The validator adds an error (correctly), but doesn't return. Execution continues to `ElementsAs`,
  which silently produces an empty slice for an unknown list. The next index access (`retentionsObjects[0].As(...)`) panics.

  ### Bug 2 — length-check branch also doesn't `return`

  ```go
  if length := len(retentionsObjects); length != 4 {
      resp.Diagnostics.AddError(...)
  }   // ← falls through to retentionsObjects[0].As(...)
  ```

  Same shape: error added, execution continues, index access panics when the slice is empty (or wrong-length). Either bug alone could cause the panic; both are present in the same function.

  ## Why was it introduced

  `git log -L 109,144:internal/provider/dataengine/resource_coralogix_archive_retentions.go` traces the validator to commit [`0121394` "Terraform to use coralogix sdk
  (#296)"](https://github.com/coralogix/terraform-provider-coralogix/commit/0121394) — a broad SDK-conversion commit, not validation-focused. The "AddError without return" pattern looks like an incomplete implementation rather than
  intentional behaviour; no documented design rationale.

  ## Is this a breaking change?

  **No.**

  - Null/unknown handling: the validator now skips silently, matching the Plugin Framework convention. The schema's existing `Required: true` (line 97) handles genuine "user didn't set it" at apply time, so users who legitimately omit
  `retentions` still get the framework's standard required-attribute error.
  - Length-check return: same error message as before; just prevents the subsequent panic. Users with the wrong number of retentions still see `retentions list must have 4 elements but got N`.

  There's also a redundant `listvalidator.SizeBetween(4, 4)` already on the schema (line 99) that catches wrong-length lists earlier, so the validator's length check rarely fires anyway.

  ## Risks

  - **Required-field enforcement now relies entirely on the schema flag.** Mitigated: that's the framework's documented pattern and is already how every other Required attribute in this provider behaves. Verified by repro: omitting
  retentions still produces a clear required-attribute error.
  - **Variable path was previously untested.** Mitigated by the new acceptance test (see Test plan).

  ## Changes

  | File | Change |
  |---|---|
  | `internal/provider/dataengine/resource_coralogix_archive_retentions.go` | (1) Replace `AddError` with early `return` for null/unknown `ConfigValue`. (2) Add `return` after the length-check error to prevent the subsequent index panic. |
  | `internal/provider/resource_coralogix_archive_retentions_test.go` | Add `TestAccCoralogixResourceArchiveRetentions_via_variable` — supplies retentions through a Terraform variable and `ConfigVariables` (the test-framework analogue of `.tfvars` / `TF_VAR_*`). Introduces `terraform-plugin-testing/config` import (first use in this repo). |

  ## Test plan

  - [x] Manually reproduced the panic on master (`make install` from clean master + variable-supplied config → panic at `resource_coralogix_archive_retentions.go:136`).
  - [x] Confirmed fix locally end-to-end against EU2: `terraform plan` succeeds with the dev-override warning, shows 4 retentions to be created, no panic.
  - [x] `TestAccCoralogixResourceArchiveRetentions_via_variable` — new acceptance test exercising the variable path (the path that broke).
  - [x] `TestAccCoralogixResourceResourceArchiveRetentions` — existing literal-path test continues to pass.

  ### Acceptance tests

  ```
  $ make testacc TESTARGS='-run=TestAccCoralogixResourceArchiveRetentions'

  (to be populated)
  ```

  ### Release Note

  ```release-note
  fix: prevent panic in `coralogix_archive_retentions` when `retentions` is supplied via a Terraform variable (BUGV2-5395)
  ```

  ### References

  - [BUGV2-5395](https://coralogix.atlassian.net/browse/BUGV2-5395) — original bug report with full panic trace and reproduction.
  - Commit [`0121394` (#296)](https://github.com/coralogix/terraform-provider-coralogix/pull/296) — introduced the validator.

  ### Community Note

  Click the copy button on that block (or select-all + copy). The 3-backtick fences inside it stay intact.

[BUGV2-5395]: https://coralogix.atlassian.net/browse/BUGV2-5395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BUGV2-5395]: https://coralogix.atlassian.net/browse/BUGV2-5395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/dataengine/resource_coralogix_archive_retentions.go
+++ b/internal/provider/dataengine/resource_coralogix_archive_retentions.go
@@ -118,7 +118,7 @@ func (r retentionsValidator) MarkdownDescription(_ context.Context) string {
 
 func (r retentionsValidator) ValidateList(ctx context.Context, req validator.ListRequest, resp *validator.ListResponse) {
 	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
-		resp.Diagnostics.AddError("error on validating retentions", "retentions can not be null or unknown")
+		return
 	}
 
 	var retentionsObjects []types.Object
@@ -130,6 +130,7 @@ func (r retentionsValidator) ValidateList(ctx context.Context, req validator.Lis
 
 	if length := len(retentionsObjects); length != 4 {
 		resp.Diagnostics.AddError("error on validating retentions", fmt.Sprintf("retentions list must have 4 elements but got %d", length))
+		return
 	}
 
 	var archiveRetentionResourceModel ArchiveRetentionResourceModel
--- a/internal/provider/resource_coralogix_archive_retentions_test.go
+++ b/internal/provider/resource_coralogix_archive_retentions_test.go
@@ -17,6 +17,7 @@ package provider
 import (
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-testing/config"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 )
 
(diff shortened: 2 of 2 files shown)


internal/provider/dataengine/resource_coralogix_archive_retentions.go
```diff
@@ -118,7 +118,7 @@ func (r retentionsValidator) MarkdownDescription(_ context.Context) string {
 
 func (r retentionsValidator) ValidateList(ctx context.Context, req validator.ListRequest, resp *validator.ListResponse) {
 	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
-		resp.Diagnostics.AddError("error on validating retentions", "retentions can not be null or unknown")
+		return
```
povilasv, line 121: Should we keep:
		resp.Diagnostics.AddError("error on validating retentions", "retentions can not be null or unknown")
and add return?
?

dmenilo1: The AddError. dropped for two reasons:
1. Plugin Framework convention. Validators should skip null/unknown rather than report them as errors.
    The framework calls ValidateResourceConfig during the plan phase even when the value is unknown (e.g.,    when it comes from a variable or an unresolved resource output). The schema's existing Required: true (line   97) handles "user didn't set it" at apply time; the validator doesn't need to.
2. Empirical impact on the bug repro. With the suggested form (keep AddError, add return), the variable path would still be broken — instead of panicking, plan would fail with "retentions can not be null or unknown". The bug's whole point is that retentions = var.retentions_list is a valid config; it just arrives as unknown at validate time when supplied via a variable. The user would see a misleading error.

povilasv (approved):

PR #518, opened 2026-05-05
fix(BUGV2-1291): accept observation_field in logs aggregation validator

dmenilo1:
### Description

`logsAggregationValidator` rejects any logs-aggregation block that uses `observation_field` instead of `field`, even though the schema declares both as `Optional` and `ExpandLogsAggregation` already passes either to the SDK. Configurations like

```hcl
aggregation = {
  type = "avg"
  observation_field = {
    keypath = ["meta", "responseTime"]
    scope   = "user_data"
  }
}
```

fail at plan time with `logs aggregation validation failed: when type is 'avg', 'field' must be set`. That's a self-inconsistency — the validator forbids what the rest of the provider accepts.

The behaviour change in `logsAggregationValidator.ValidateObject`:

| `type` | Before | After |
|---|---|---|
| `count` | error if `field` is set | error if `field` *or* `observation_field` is set |
| non-count | error if `field` is null | error only if **both** `field` and `observation_field` are null |
| any non-count | (no check) | error if both are set (mutually exclusive) |
| `percentile` | unchanged | unchanged |

Error messages now name `observation_field` so users see the alternative.

### Customer context
The customer in BUGV2-1291 hit drift where the API echoed `field = "meta.responseTime.numeric"` back as `observation_field = {keypath = ["meta", "responseTime"], scope = "user_data"}` (segment dropped). They tried to fix the drift by switching their config to `observation_field`, but the validator blocked the attempt.

This PR is intentionally scoped to the validator. The drift root cause is server-side (dashboards-api dropping the `.numeric` segment) and is being followed up with that team separately.

### Acceptance tests
  - Added a unit test for the validator (`internal/provider/dashboards/dashboard_widgets/common_test.go`, 12 cases). A full acceptance test is intentionally not added — it would entangle with the server-side round-trip drift this PR does not fix and would flake when the backend is corrected.
  - Unit tests pass:
    ```
    $ go test ./internal/provider/dashboards/dashboard_widgets/ -run TestLogsAggregationValidator -v
    ...
    --- PASS: TestLogsAggregationValidator (0.00s)
    ```
  - Local `terraform plan` against EU2 with `observation_field` for an `avg` aggregation succeeds where it previously errored.

### Release Note
```release-note
fix(dashboards): allow `observation_field` as an alternative to `field` in logs aggregations (BUGV2-1291)
```

### DIFF
--- a/internal/provider/dashboards/dashboard_widgets/common.go
+++ b/internal/provider/dashboards/dashboard_widgets/common.go
@@ -850,10 +850,19 @@ func (l logsAggregationValidator) ValidateObject(ctx context.Context, req valida
 	}
 
 	aggregationType := aggregation.Type.ValueString()
-	if aggregationType == "count" && !aggregation.Field.IsNull() {
-		resp.Diagnostics.Append(diag.NewErrorDiagnostic("logs aggregation validation failed", "when type is `count`, `field` cannot be set"))
-	} else if aggregationType != "count" && aggregation.Field.IsNull() {
-		resp.Diagnostics.Append(diag.NewErrorDiagnostic("logs aggregation validation failed", fmt.Sprintf("when type is `%s`, `field` must be set", aggregationType)))
+	fieldSet := !aggregation.Field.IsNull()
+	observationFieldSet := !aggregation.ObservationField.IsNull()
+
+	if aggregationType == "count" {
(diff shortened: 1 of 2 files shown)


dmenilo1:
@codex review

assafad1 (approved):

PR #522, opened 2026-05-07
fix(BUGV2-6012): stop coralogix_parsing_rules.Update from duplicating rule groups

assafad1:
## Summary

Fix `coralogix_parsing_rules.Update` calling the **Create** endpoint instead of **Update**, which silently orphaned the previous rule group at Coralogix on every update and was the source of recurring parsing-rule quota exhaustion in the test account.

## Root Cause

`internal/provider/parsing_rules/resource_coralogix_parsing_rules.go:Update` invoked `RuleGroupsServiceCreateRuleGroup(ctx)` with the plan body, then `flattenParsingRules(result.RuleGroup)` and `resp.State.Set(...)`. Each Update therefore:

1. Created a brand-new rule group at Coralogix with a brand-new ID.
2. Replaced Terraform state's ID with the new one — losing all reference to the original.
3. Left the original rule group orphaned at the API. `terraform destroy` later only deleted whatever ID state currently held, never the orphans.

This affected:
- **Real users** running `terraform apply` after editing a `coralogix_parsing_rules` resource — every diff produced a permanent orphan.
- **CI** — `TestAccCoralogixResourceParsingRules_update` and `TestAccCoralogixResourceParsingRules_update_order_inside_rule_group` each leaked one rule per acceptance run (cron + every PR), accumulating against the per-account parsing-rule quota faster than the nightly `cleanup.yml` could sweep.

The pinned SDK (`coralogix-management-sdk@v1.9.4-…`) exposes `RuleGroupsServiceUpdateRuleGroup(ctx, groupId)` with the same `RuleGroupsServiceCreateRuleGroupRequest` body and an `UpdateRuleGroupResponse` whose `RuleGroup` field has identical shape — so the fix is a one-line endpoint swap that carries the existing `plan.ID` through.

The companion test bug in `testAccCheckParsingRuleDestroy` (unguarded `*resp.RuleGroup.Id` deref when the GET returns success with a nil `RuleGroup`) couldn't have caught this leak — orphaned rules aren't in TF state, so CheckDestroy never iterates them — but it could mask future regressions where the leak *is* visible to state, so it's hardened in the same change.

## Why was it introduced

Introduced in #447 ("OpenAPI SDK II", commit 3ec543f, 2025-12-02) when the framework `coralogix_parsing_rules` resource was first added against the OpenAPI-generated SDK. The Update implementation appears to have been copy-pasted from Create without swapping the endpoint — the existing error string `"Error replacing coralogix_parsing_rules"` and the `"Create"` operation tag passed to `FormatOpenAPIErrors` are tells.

## Breaking change?

**No.** The change moves Update from `POST /create` to `PUT /update` against the same backend service with the same request body. Externally observable schema, attributes, and import behaviour are unchanged. State migration is a no-op: Terraform state already holds the (latest) rule's ID, which is what the new Update path uses.

Note for users with existing pre-fix orphans: this PR does not retroactively delete them. They remain unreferenced at Coralogix and should be cleaned up via the nightly `scripts/cx-test-resource-cleanup.go` script (test envs) or manually via the Coralogix UI / SDK in production envs.

## Risks

- **Update endpoint divergence at the API level** — mitigated: SDK exposes both endpoints with identical request/response body shapes, so request marshalling is unchanged and the existing `flattenParsingRules` continues to work. Verified against `UpdateRuleGroupResponse.RuleGroup *RuleGroup` (same as Create).
- **`plan.ID` empty during Update** — mitigated: `id` is `Computed` in the schema and Terraform preserves Computed values in plan during an in-place update; identical pattern is already used by `apm/resource_coralogix_slo.go:485`.
- **CheckDestroy guard hiding genuine "rule still exists"** — mitigated: nil-guards only short-circuit on responses that previously panicked; if the GET succeeds with a populated `RuleGroup.Id` matching `rs.Primary.ID`, the existing `fmt.Errorf("RuleGroup still exists: %s", …)` still fires.

## Changes

| File | Change |
| --- | --- |
| `internal/provider/parsing_rules/resource_coralogix_parsing_rules.go` | `Update` now calls `RuleGroupsServiceUpdateRuleGroup(ctx, plan.ID)` instead of `RuleGroupsServiceCreateRuleGroup(ctx)`. Error text/tag corrected to `Update`. |
| `internal/provider/resource_coralogix_parsing_rules_test.go` | Added `captureParsingRuleID` + `assertParsingRuleIDUnchanged` helpers; wired them into `TestAccCoralogixResourceParsingRules_update` and `TestAccCoralogixResourceParsingRules_update_order_inside_rule_group` so any future Update implementation that recreates instead of updating fails the test. Hardened `testAccCheckParsingRuleDestroy` against nil deref. |

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `make test` — clean
- [ ] `make testacc TESTARGS='-run=TestAccCoralogixResourceParsingRules_update$'` — must pass post-fix; without the fix, the new `assertParsingRuleIDUnchanged` would fail with `RuleGroup ID changed across update: <old> -> <new> (resource was recreated, leaking the original)`.
- [ ] `make testacc TESTARGS='-run=TestAccCoralogixResourceParsingRules_update_order_inside_rule_group'` — same, for the second update test.
- [ ] `make testacc TESTARGS='-run=TestAccCoralogixResourceParsingRules'` — full parsing-rules suite, no regressions.
- [ ] Manual leak check: list parsing rule groups in the test account before and after the two update tests; expect `0` net new rules post-fix (was `+2`).

### Acceptance tests
- [ ] Have you run the acceptance tests on this branch? *(pending — needs `CORALOGIX_API_KEY` / `CORALOGIX_ENV` against the test account)*

\`\`\`
$ make testacc TESTARGS='-run=TestAccCoralogixResourceParsingRules_update'

(pending — to be filled in once acceptance tests are run against the test environment)
\`\`\`

### Release Note
\`\`\`release-note
resource/coralogix_parsing_rules: fix \`Update\` recreating the rule group at the backend (new ID, orphaning the previous rule). \`terraform apply\` now updates in place; the rule's ID is stable across edits.
\`\`\`

### References

- Jira: [BUGV2-6012](https://coralogix.atlassian.net/browse/BUGV2-6012)
- Bug introduced in #447 (OpenAPI SDK II, commit 3ec543f).
- Branch was originally created under the working hypothesis that test cleanup was at fault; root cause turned out to be the resource itself, but the test-side hardening (CheckDestroy nil-guard + ID-stability assertions) is included to prevent silent regressions.

### Community Note

[BUGV2-6012]: https://coralogix.atlassian.net/browse/BUGV2-6012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/provider/parsing_rules/resource_coralogix_parsing_rules.go
+++ b/internal/provider/parsing_rules/resource_coralogix_parsing_rules.go
@@ -441,15 +441,16 @@ func (r *ParsingRulesResource) Update(ctx context.Context, req resource.UpdateRe
 		return
 	}
 
+	id := plan.ID.ValueString()
 	rq := extractParsingRules(plan)
 
 	result, httpResponse, err := r.client.
-		RuleGroupsServiceCreateRuleGroup(ctx).
+		RuleGroupsServiceUpdateRuleGroup(ctx, id).
 		RuleGroupsServiceCreateRuleGroupRequest(*rq).
 		Execute()
 	if err != nil {
-		resp.Diagnostics.AddError("Error replacing coralogix_parsing_rules",
-			utils.FormatOpenAPIErrors(cxsdkOpenapi.NewAPIError(httpResponse, err), "Create", rq),
+		resp.Diagnostics.AddError("Error updating coralogix_parsing_rules",
+			utils.FormatOpenAPIErrors(cxsdkOpenapi.NewAPIError(httpResponse, err), "Update", rq),
 		)
 		return
 	}
--- a/internal/provider/parsing_rules/resource_coralogix_parsing_rules.go
+++ b/internal/provider/parsing_rules/resource_coralogix_parsing_rules.go
@@ -34,8 +34,10 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/setdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
@@ -260,8 +262,10 @@ func (r *ParsingRulesResource) Schema(_ context.Context, _ resource.SchemaReques
 		Version: 0,
 		Attributes: map[string]schema.Attribute{
 			"id": schema.StringAttribute{
-				Optional: true,
 				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 			"name": schema.StringAttribute{
 				Required:    true,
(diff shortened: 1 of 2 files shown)


assafad1:
@codex review

povilasv (approved):

PR #560, opened 2026-06-11
[TEST] fix(events2metric): return after failed create to avoid adopting an unrelated metric

dmenilo1:
*What*

coralogix_events2metric could overwrite an internal events2metric after a failed create.

When CreateE2M returns an error, Create recorded the diagnostic but did not return. Execution fell through to flattenE2M with the nil create response, producing a model whose id is the empty string, which Terraform then persisted. On the next refresh, Read sent that empty id to GetE2M, the backend answered with an arbitrary (often internal) events2metric, and the subsequent plan proposed destroying and replacing it.

*Fix*

- Add the missing return after the create-error diagnostic in Create, so a failed create no longer writes empty-ID state.
- In Read, remove the resource from state when its stored id is empty rather than fetching an unrelated events2metric.

*Test*

Adds TestAccCoralogixResourceEvents2MetricCreateConflict: creates one events2metric, then attempts a second that reuses the same generated metric name (rejected by the API), asserts the create error surfaces, and re-plans the first resource expecting an empty plan. Red on the pre-fix code, green with this change.

*Validation*

No single-apply reproducer was attached at triage: the bug needs a metric-name collision plus a two-step apply, so a single terraform apply cannot surface it. The acceptance test above encodes the scenario instead; live TF_ACC verification against a tenant is deferred.

TEST-MODE pipeline-validation PR. Draft, not for merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+#### resource/coralogix_events2metric
+
+- FIX: A failed create no longer writes an empty-ID resource into Terraform state. The empty ID caused the next refresh to fetch an arbitrary Events2Metric (typically an internal one), which the subsequent plan would try to destroy and replace. Refresh now also drops legacy empty-ID entries from state instead of adopting an unrelated Events2Metric.
+
 #### resource/coralogix_alert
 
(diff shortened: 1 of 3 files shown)


dmenilo1:
Closing — this was a pipeline-validation test PR (triage→PR-work dry run on an already-fixed issue). The fix it reproduced is already on master via #555. The Read empty-ID guard it surfaced is worth shipping as a real follow-up; tracking separately.

PR #565, opened 2026-06-15
docs(tco_policies_logs): document priority as the quota-based fallback

dmenilo1:
## What
Documents that, for a quota-based TCO log policy (when `quota_based_priority_override` is set), the policy's top-level **`priority`** is the **"Route the remaining quota to"** fallback — the priority applied once all `usage_tiers` are exhausted — and that it must be more restrictive than the last tier (most→least restrictive: `block`, `low`, `medium`, `high`).

## Why
Users had no way to know how to set the UI's "Route the remaining quota to" value in Terraform; the attribute that controls it (`priority`) was documented only generically, and the quota fallback mapping was undocumented. This led to a support ticket where a user added a non-existent setting and set `priority` equal to the last tier, hitting the "fallback must be more restrictive than the last tier" error.

## How
Schema-only change: updated the `MarkdownDescription` for `priority` and `quota_based_priority_override` in `resource_coralogix_tco_policies_logs.go`. Docs regenerated via `make generate` (tfplugindocs) — `docs/resources/tco_policies_logs.md` and `docs/data-sources/tco_policies_logs.md`. No behavior change.

Refs CDS-3036

### DIFF
--- a/docs/data-sources/tco_policies_logs.md
+++ b/docs/data-sources/tco_policies_logs.md
@@ -39,8 +39,8 @@ Read-Only:
 - `id` (String) tco-policy ID.
 - `name` (String) tco-policy name.
 - `order` (Number) The policy's order between the other policies.
-- `priority` (String) The policy priority. Can be one of ["block" "high" "low" "medium"].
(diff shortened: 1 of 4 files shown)


dmenilo1:
@codex review

vjofusha (approved):

PR #603, opened 2026-07-15
Migrate Dashboards to OpenAPI REST client

markshleifer-coralogix:
## Description

This PR migrates `coralogix_dashboard` and its data source from the legacy gRPC dashboard client to the generated OpenAPI REST client. It keeps the existing Terraform schema and state compatible while moving CRUD, reads, imports, and `content_json` handling onto the REST dashboard models.

## Change size

- Across all added and modified files, tests account for roughly **9.3k added lines**, while the production implementation accounts for roughly **2.9k added lines** (outdated numbers, but this has roughly remained so). Most of the PR size is therefore migration coverage rather than runtime code.

## Implementation review

- Replaces the dashboard client in `ClientSet` and routes resource/data-source operations through a small REST adapter with consistent error handling, request IDs, not-found behavior, and idempotent deletes.
- Converts dashboard expand/flatten logic from protobuf types to generated OpenAPI models, including their nested `oneOf` representations across widgets, queries, filters, variables, annotations, layouts, and time frames.
- Preserves compatibility for existing state and configurations: prior schema versions are upgraded, `content_json` accepts both protobuf-style `snake_case` and REST `lowerCamelCase`, and unknown OpenAPI properties are discarded before writes as they were with the old protobuf path.
- Covers failure and partial-create paths so failed reads or updates do not poison state, and preserves access-policy behavior through create, read, update, import, and data-source flows.
- Also adds horizontal-bar DataPrime query support and fixes REST model mismatches in data-table spans aggregation and hexagon spans aggregation.

## E2E coverage

- Widget and query matrix: Covers the supported widget types—including line charts, data tables, gauges, bar charts, horizontal bar charts, and hexagons—with logs, metrics, spans, and DataPrime query variants.
- oneOf permutations: Exercises nested REST oneOf branches across queries, aggregations, filters, variables, annotations, time frames, resolutions, axes, and presentation options.
- Transitions and clearing: Verifies switching between oneOf branches during updates, removing optional values, and applying again without state drift.
- Full Terraform lifecycle: Covers create, read, update, refresh, import, data-source reads, and second-plan stability for both structured dashboards and content_json.
- Backend-created dashboards: Creates dashboards directly through REST and verifies that Terraform can import or read them, including backend-generated IDs and values.
- content_json compatibility: Tests both legacy protobuf snake_case and REST lowerCamelCase fields, folder overrides, dynamic query tables, and unknown fields.
- Failure behavior: Covers rejected creates and updates, failed post-create reads, cleanup failures, missing dashboards, transient read errors, and retryable deletes to ensure Terraform state remains usable.
- Migration: Runs dedicated E2E tests that create dashboards with the pinned legacy gRPC provider and then manage them with the current REST implementation. It also covers upgrades from older Terraform dashboard schema state.
## Faster acceptance-test execution

Dashboard acceptance tests now run separately from the rest of the provider and are split into eight parallel shards, with legacy-to-REST migration tests in their own runner. This reduces overall CI wall-clock time while keeping the broader dashboard matrix isolated and easier to diagnose.
The overall time it takes to run is roughly 6m, as compared to 16-20 minutes on master.

## Manual testing
Tested the following scenarios by applying, checking empty plan after apply, then making a change, and finally destroying.

1. Empty Dashboard
2. One branch selected at each meaningful oneOf
3. Two branches selected in one oneOf: reject before API mutation
4. Meaningful oneOf transitions on the same dashboard
5. One meaningful permutation per widget family, including updates for each widget
6. Structured upgrade from released 3.7.0 (gRPC) to this branch (REST)
A dashboard with a data prime query type in a horizontal bar
7. Dashboard with `content_json` exported from the UI. Tested widget types: dynamic widget, geo map, stat card, line chart, data table, gauge, poly stat, heat map, pie chart, vertical bar chart, horizontal bar chart, markdown, data prime. Also tested: annotations, variables and filters.
8. Create the dashboard from point 7 in version 3.7.0 (gRPC), upgrade, see that plan is empty, make a change and see it works.

### Release Note
```
- CHORE: Migrate dashboard operations from the legacy gRPC client to the REST client.
- FEAT: Add support for DataPrime queries in `horizontal_bar_chart` widgets.
```


internal/clientset/clientset.go
```diff
@@ -267,6 +266,7 @@ func NewClientSet(region string, apiKey string, grpcTarget string) *ClientSet {
 		archiveMetrics:        cs.ArchiveMetrics(),
 		alerts:                cs.Alerts(),
 		archiveRetentions:     cs.ArchiveRetentions(),
+		dashboardsOpenapi:     cs.Dashboards(),
```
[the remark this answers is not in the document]
markshleifer-coralogix: Dashboards are not the first to use the OpenAPI client so they will get the same global configuration.

internal/provider/dashboards/dashboardjson/unmarshal.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		return err
+	}
+
+	normalized := normalizeProtoFieldNames(raw, targetType)
+	encoded, err := json.Marshal(normalized)
+	if err != nil {
+		return fmt.Errorf("marshal normalized dashboard JSON: %w", err)
+	}
+	if err := json.Unmarshal(encoded, target); err != nil {
```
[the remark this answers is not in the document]
markshleifer-coralogix: This is an existing limitation that exists on master as well.

internal/clientset/clientset.go
```diff
@@ -126,7 +127,7 @@ func (c *ClientSet) DataSet() *cxsdk.DataSetClient {
 	return c.dataSet
 }
 
-func (c *ClientSet) Dashboards() *cxsdk.DashboardsClient {
+func (c *ClientSet) Dashboards() *dashboardservice.DashboardServiceAPIService {
```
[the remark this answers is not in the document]
markshleifer-coralogix: It's already broken on master so I will address it in a separate PR.

internal/provider/dashboards/dashboard_schema/v4.go
```diff
@@ -133,7 +135,7 @@ func dashboardSchemaAttributesV4() map[string]schema.Attribute {
 																					"time_frame":       dashboardwidgets.TimeFrameSchema(),
 																				},
 																				Validators: []validator.Object{
-																					objectvalidator.ExactlyOneOf(
+																					dashboardwidgets.ExactlyOneOfObject(
```
assafad1, line 138: why are these needed? is it changing anything in TF validations? are we ok with that?

internal/provider/dashboards/dashboard_schema/v4.go
```diff
@@ -989,6 +991,8 @@ func dashboardSchemaAttributesV4() map[string]schema.Attribute {
 									"selected_values": schema.ListAttribute{
 										ElementType: types.StringType,
 										Optional:    true,
+										Computed:    true,
+										Default:     listdefault.StaticValue(types.ListValueMust(types.StringType, []attr.Value{})),
```
assafad1: why are we changing that?

internal/provider/dashboards/dashboardjson/unmarshal.go
```diff
@@ -0,0 +1,163 @@
+// Copyright 2026 Coralogix Ltd.
```
assafad1, line 1: can you please explain why we needed it if the exported json looks the same? are existing JSONs not passing through the OpenAPI SDK?

internal/provider/dashboards/dashboard_openapi_client.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return formatDashboardOpenAPIError(httpResponse, err, dashboardOpenAPIOperationDelete, id)
+}
+
+func newDashboardOpenAPICreateRequest(dashboard dashboardservice.Dashboard, accessPolicy *string) dashboardservice.CreateDashboardRequestDataStructure {
+	discardOpenAPIAdditionalProperties(&dashboard)
+	request := dashboardservice.CreateDashboardRequestDataStructure{
+		Dashboard: dashboard,
+		RequestId: newDashboardOpenAPIRequestID(dashboardOpenAPIOperationCreate),
```
assafad1, line 131: why do we need to provide a request id? did we provide it before the PR?

internal/provider/dashboards/dashboard_openapi_client.go
```diff
@@ -0,0 +1,206 @@
...
+			discardAdditionalPropertiesValue(value.MapIndex(key))
+		}
+	}
+}
+
+func newDashboardOpenAPIRequestID(operation string) string {
+	return fmt.Sprintf("%s-%s-%s", dashboardOpenAPIRequestIDPrefix, operation, uuid.NewString())
+}
+
+func formatDashboardOpenAPIError(httpResponse *http.Response, err error, operation string, request any) error {
```
assafad1: don't we have shared formatting here or in the SDK we can reuse here?

internal/provider/dashboards/dashboard_openapi_client.go
```diff
@@ -0,0 +1,206 @@
...
+	if err == nil {
+		return nil
+	}
+
+	apiErr := cxsdkOpenapi.NewAPIError(httpResponse, err)
+	return fmt.Errorf("dashboard REST %s failed: %s", operation, utils.FormatOpenAPIErrors(apiErr, operation, request))
+}
+
+func isDashboardOpenAPINotFound(httpResponse *http.Response, err error) bool {
```
assafad1, line 198: don't we have SDK utils for that?

internal/provider/dashboards/dashboard_openapi_client.go
```diff
@@ -0,0 +1,206 @@
...
+
+// discardOpenAPIAdditionalProperties restores protojson's historical
+// DiscardUnknown behavior for content_json. OpenAPI Generator captures unknown
+// JSON fields in AdditionalProperties, but the protobuf HTTP endpoint rejects
+// them when they are sent back.
+func discardOpenAPIAdditionalProperties(value any) {
+	discardAdditionalPropertiesValue(reflect.ValueOf(value))
+}
+
+func discardAdditionalPropertiesValue(value reflect.Value) {
```
assafad1, line 161: why is this needed? can you please provide examples?

internal/provider/dashboards/dashboard_schema/v4.go
markshleifer-coralogix: `ExactlyOneOfObject`wraps `objectvalidator.ExactlyOneOf` and uses it internally.
I started with `ExactlyOneOf`, but when testing I saw that when using multiple oneOf branches together, they were rejected as they should, but the error message was not clear at all when the oneOf was nested, so I made a custom validator to display a more user friendly message.

internal/provider/dashboards/dashboard_schema/v4.go
markshleifer-coralogix: Good point.
That schema tweak was added because `flattenDashboardVariableSelectedValues` returned an empty list (instead of null) for the "select all" case, which caused drift. But now I see that the API treats null and an empty list the same way for "select all", so I reverted the flatten function to return null again and removed the schema change. Added a test that confirms there's no drift when selected_values is left out.

internal/provider/dashboards/dashboardjson/unmarshal.go
markshleifer-coralogix: `protojson.Unmarshal` (the old usage) accepted either the proto field name (`snake_case`, e.g. `series_count_limit`) or the JSON name (`lowerCamelCase`, e.g. `seriesCountLimit`).
Since current users that rely on the gRPC client might have used `snake_case`, this umurshal does a translation from `snake_case` to `lowerCamelCase` to be backward compatible.
I didn't find actual evidence of such usage, but I wanted to be extra protective and make sure that it's covered as well.

internal/provider/dashboards/dashboard_openapi_client.go
markshleifer-coralogix: Request id is part of the proto contract and marked as required.
The BE is probably permissive about the absence of it, but I would say that adding it wouldn't hurt since it's most likely used for idempotency as in similar dashboard requests.

internal/provider/dashboards/dashboard_openapi_client.go
markshleifer-coralogix: Good point. Changed it to use `FormatOpenAPIErrors` directly as many resources do.

internal/provider/dashboards/dashboard_openapi_client.go
markshleifer-coralogix: I didn't find a wrapper for it in the SDK. It's basically two SDK calls wrapped into one function call.

internal/provider/dashboards/dashboard_openapi_client.go
assafad1: I don't have anything against it in general, just wonder if this PR should include such changes, rather than being a refactor only. But if you say it's ok, I'm happy.

internal/provider/dashboards/dashboard_openapi_client.go
assafad1: But you still wrap it in this file. I just want to make sure the code is consistent across different resources.
WDYT?

internal/provider/dashboards/dashboard_openapi_client.go
assafad1: see https://github.com/coralogix/terraform-provider-coralogix/pull/603/changes#r3614936733

internal/provider/dashboards/dashboard_openapi_client.go
markshleifer-coralogix: This is for backward compatibility of `content_json`.
Before the OpenAPI migration, `content_json` was parsed with `protojson.UnmarshalOptions{DiscardUnknown: true}` (`dashboard_schema/common.go`: https://github.com/coralogix/terraform-provider-coralogix/blob/master/internal/provider/dashboards/dashboard_schema/common.go#L79), so unknown/legacy fields in existing dashboards were silently dropped. The OpenAPI-generated client instead captures those same unrecognized fields into AdditionalProperties, and if sent back as-is, the HTTP endpoint rejects the request. `discardOpenAPIAdditionalProperties` strips them before Create/Replace so upgrading users with older dashboards don't start seeing errors on fields they never touched.
Other resource don't have this complexity since they don't have `content_json`.

internal/provider/dashboards/dashboard_openapi_client.go
markshleifer-coralogix: I removed this wrapper in a later commit: https://github.com/coralogix/terraform-provider-coralogix/pull/603/changes/90655a28d567f73d82fff41d0e4d8ecfde251344

internal/provider/dashboards/dashboard_openapi_client.go
markshleifer-coralogix: Didn't find a usage of `cxsdkOpenapi.IsNotFound` in this repo, so there is no reference point. A possible change would be a one-line of `cxsdkOpenapi.IsNotFound(cxsdkOpenapi.NewAPIError(httpResponse, err))`, but IMO it's pretty much the same as now.

internal/provider/dashboards/dashboard_schema/v4.go
markshleifer-coralogix: I have actually reverted it because it fixed a bug that was on master as well.
When the dashboard had `selected_values = []` in the definition, applying it failed since there was a mismatch between the state (which had `selection.All`) and how the flattening code treated the `selection.All` value the API returns.
The error from master:
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to coralogix_dashboard.test, provider
│ "provider[\"registry.terraform.io/coralogix/coralogix\"]" produced an unexpected new
│ value: .variables[0].definition.multi_select.selected_values: was
│ cty.ListValEmpty(cty.String), but now null.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue
│ tracker.
╵
```

assafad1 (approved):

PR #609, opened 2026-07-22
perf(dashboards): replace per-child oneof validators with cheaper parent-attached ExactlyOneOfChildren

markshleifer-coralogix:
### Description

`coralogix_dashboard` plans get slow as a dashboard accumulates widgets, with `ValidateResourceConfig` dominating plan time on large dashboards. This PR removes that hotspot for the V4 schema (the schema every live resource instance actually uses — V1-V3 are legacy, reachable only once per resource via `StateUpgraders`).

Perf results, measured with `time terraform plan` against a fixture of 20 dashboards covering a variety of widget setups (same fixture before/after):

| | Before | After |
|---|---|---|
| `terraform plan` (20 dashboards with widgets) | **4m11s** | **15.277s** |

~16.5x wall-clock improvement.

**Root cause:** every `oneof`-style group in the schema (widget type selection, `logs`/`metrics`/`spans`/`data_prime` query unions, time frames, annotation sources, folder `id`/`path`, ...) attached an `ExactlyOneOf`-style validator to **each child** in the group, referencing its siblings by `path.Expression`. Resolving those expressions means walking the whole config tree via `config.PathMatches` — once per sibling, per group, per widget instance. That cost repeats for every widget, so plan time scales with widget count.

This PR fixes it with two changes, bundled because the second only pays off once the first is done:

1. **Attach once per group, not once per child.** New `dashboardwidgets.ExactlyOneOfChildren(childNames...)` is attached to the *parent* object of a oneof group instead of to each child. A group of N siblings now carries one validator invocation instead of N.
2. **Make that one validator itself cheap.** `ExactlyOneOfChildren` reads `req.ConfigValue.Attributes()` directly — the parent object it's attached to already has all its children's values in hand — instead of resolving `path.Expression`s against the full config via `config.PathMatches`/`config.GetAttribute`. This replaces an O(config size) tree walk with an O(children-in-the-group) map lookup.

The widget-type oneof (`SupportedWidgetsValidatorWithout`, an 8-way check previously attached to all 8 widget-type siblings) collapses the same way into a single `SupportedWidgetsExactlyOneOfChildren()` attached once to the `definition` object.

The old `ExactlyOneOfObject`/`ExactlyOneOfString`/`ExactlyOneOfInt64`/`SupportedWidgetsValidatorWithout` family is left in place (still used to gate `AddError` message rewriting on invalid configs) but is no longer wired into any V4 oneof group — every group in `v4.go`, `common.go`, `hexagon.go`, `linechart.go`, `datatable.go`, and `schema.go`'s shared builders (`TimeFrameSchema`, `FiltersSourceSchema`) now uses the parent-attached form.

#### A subtlety: unknown values must defer, not reject

`ExactlyOneOfChildren` inspects each child's concrete null/unknown/set state directly instead of delegating to the framework's stock `ExactlyOneOf`, so it has to replicate the unknown-handling itself: two or more known-and-set children is always a conflict, but if any sibling is still unknown (e.g. derived from another resource not yet resolved), it defers rather than reporting "none configured" — that sibling could still resolve to satisfy the constraint. Only once every sibling is known does "zero set" become a real error.

This is safe because `ValidateResourceConfig` isn't a one-shot check: it re-runs on every `plan`, and again as part of `apply`'s internal re-plan of each resource right before `Create`/`Update`. A first `plan` may see the value unknown and defer; by the time `apply` reaches this resource its dependencies are already created, so the value is known and the validator re-evaluates for real — rejecting it then if it's genuinely invalid. Deferring here doesn't skip validation, it just defers to the pass where the answer is actually knowable.

Covered by `TestExactlyOneOfChildrenDefersWhenSiblingUnknown` and `TestHexagonQueryExactlyOneOfChildren`.

### Profiling Analysis
Ran `pprof` on both master and this branch, using the 20 dashboards example:
- In master, `fromtftypes.AttributePath` cost 156.68s total, and 94.85% of that (148.61s) came from one specific caller: `fwschemadata.Data.PathMatches.func1` — that's the Config.PathMatches call inside the old ExactlyOneOf validators, walking the entire config tree and rebuilding nested type maps from scratch on every check.
- In this branch, that same caller (`PathMatches.func1`) only accounts for 3.32s — a 97.8% reduction in absolute terms. It's no longer the dominant caller of that code path; the majority of what's left (`NullifyCollectionBlocks.func1`, `TransformDefaults.func1`, `ReifyNullCollectionBlocks.func1`) is generic terraform-plugin-framework machinery that runs on every plan/apply regardless of validators.
- Downstream effect confirmed too. The GC/scheduler noise that dominated the master profile (`runtime.usleep` 149s, `pthread_cond_wait` 134s, `pthread_kill` 54s — together over half of all CPU) shrank proportionally in the perf profile (14s, 14s, 5.6s). The tree-walk/type-rebuild churn was generating massive allocation pressure, and removing it collapsed the GC overhead along with it, not just the walk itself.

### Test coverage

- Unit tests on `exactlyOneOfChildrenValidator` itself (`dashboard_widgets/common_test.go`): set-count matrix (0/1/2+), unknown-sibling deferral, widget-type `childNames` coverage, and the same matrix exercised through a real widget's query union.
- New schema-wiring tests (`dashboard_schema/v4_oneof_wiring_test.go`) walk the live `V4()` schema to catch partially-migrated groups (old validator left on a sibling) and `childNames` drift from the real attribute map, for both widget-level and dashboard-level groups.
- New full-schema validator tests (`dashboard_oneof_validation_test.go`) invoke real V4 attribute validators directly across widget definitions, gauge, folder, annotations, and per-element `data_prime.filters`.
- Existing rejection/round-trip unit tests and the full acceptance suite (nested-oneof, openapi, transition) are unchanged and still pass — the migration is behavior-preserving; only plan-time cost changed.
- No test measures validator performance; the numbers above are from a manual `time terraform plan` run, not CI-gated.

### Acceptance tests
- [x] Existing dashboard acceptance suite (`resource_coralogix_dashboard_nested_oneof_test.go`, `resource_coralogix_dashboard_openapi_test.go`, `resource_coralogix_dashboard_openapi_transition_test.go`, etc.) covers the oneof branches this PR rewires; no new acceptance test was added since this is a validator-internals change, not a schema-shape change.

### Release Note
```release-note
PERF: Replace per-child oneof validators (e.g. `instant`/`range`, widget type selection) with a single cheaper validator attached to the parent object, reducing plan/apply time for dashboards with many widgets.
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,3 +25,23 @@ jobs:
           go-version: ${{ matrix.go-version }}
       - name: Run build
         run: make build
+
+  test:
+    name: Unit tests
+    permissions:
+      contents: read
+    env:
+      GO111MODULE: on
+    strategy:
+      matrix:
+        go-version: [ 1.24.x ]
+        os: [ ubuntu-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Go
+        uses: actions/setup-go@37335c7bb261b353407cff977110895fa0b4f7d8
+        with:
+          go-version: ${{ matrix.go-version }}
+      - name: Run unit tests
+        run: make test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - CHORE: Migrate dashboard operations from the legacy gRPC client to the REST client.
 - FEAT: Add support for DataPrime queries in `horizontal_bar_chart` widgets.
+- PERF: Replace per-child oneof validators (e.g. `instant`/`range`, widget type selection) with a single cheaper validator attached to the parent object, reducing plan/apply time for dashboards with many widgets.
 
 # Release 3.7.0
 
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,6 @@ install: build
 	mv ${BINARY} ~/.terraform.d/plugins/${HOSTNAME}/${NAMESPACE}/${NAME}/${VERSION}/${OS_ARCH}
 
 test:
-	go test ${BUILD_ARGS} -i $(TEST) || exit 1
 	echo $(TEST) | xargs -t -n4 go test ${BUILD_ARGS} $(TESTARGS) -timeout=30s -parallel=4
 
 testacc:
--- a/internal/provider/dashboards/dashboard_oneof_validation_test.go
+++ b/internal/provider/dashboards/dashboard_oneof_validation_test.go
@@ -22,6 +22,7 @@ import (
 
 	dashboardschema "github.com/coralogix/terraform-provider-coralogix/internal/provider/dashboards/dashboard_schema"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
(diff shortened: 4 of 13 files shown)


markshleifer-coralogix:
@codex review

markshleifer-coralogix:
@codex review

markshleifer-coralogix:
@codex review

markshleifer-coralogix:
@codex review

povilasv (approved):

PR #631, opened 2026-08-06
feat(dashboards): add cross-dashboard widget reference

markshleifer-coralogix:
### Description

Adds support for cross-dashboard widget reuse on `coralogix_dashboard`. A widget can set `reference` with `dashboard_id` and `widget_id` instead of an inline `definition`. Exactly one of those two blocks is required. Local `title` and `description` on the reference widget are supported and round-trip without drift.

### Release Note
```
FEAT: Add `reference` on widgets to reuse a widget from another dashboard by `dashboard_id` and `widget_id`. Exactly one of `definition` or `reference` must be set.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 #### resource/coralogix_dashboard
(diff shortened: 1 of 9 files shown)


markshleifer-coralogix:
@codex review

assafad1 (approved):

PR #641, opened 2026-08-12
feat(dashboards): add the typed dynamic widget with the stat visualization (CX-50151)

dmenilo1:
## What this does

The dashboards API has a widget type, `dynamic`, that the provider never supported. Worse, **reading** a dashboard that already contained one failed outright — so `terraform import` and the data source broke on any dashboard built in the Coralogix UI or through `content_json`.

This adds `dynamic` to HCL: the whole query side, and one visualization, `stat`.

```hcl
definition = {
  dynamic = {
    query_definitions = [{
      query = { logs = { lucene_query = "*", aggregations = [{ type = "count" }] } }
    }]
    visualization = { stat = { unit = "bytes", threshold_type = "absolute" } }
  }
}
```

A visualization that isn't supported yet still fails the read with a clear message telling the user to use `content_json`; it never writes half a widget into state.

## Where to look

| What | Where |
|---|---|
| The read that used to fail | `resource_coralogix_dashboard.go`, the flatten dispatch's `dynamic` branch |
| What the user writes | `dynamic.go` → `DynamicSchema()` |
| The two directions that must mirror each other | `expandDynamicVisualization` / `flattenDynamicVisualization` |

**The thing worth checking:** the write direction (HCL → API) and the read direction (API → state) have to carry the same fields. If the read drops a field the write sends, `import` loses it and the next `apply` sends it back empty — **clearing the setting on the server**. That's why `stat` models all of its API fields instead of a useful subset.

Adding a variant means wiring it into five places: the schema map, the attribute-type map, both dispatch switches, and the exactly-one-of list. Miss one and you get either an error on read or a branch that silently never fires.

Two renames, both to match what this provider already does elsewhere. Nothing else deviates.

**1. `aggregations`, not `aggregation`**
- *What it is:* the calculations a query runs — count, average, and so on.
- *API (wire):* `"aggregation": [ { "type": "count" } ]` — singular name, holds a list.
- *Terraform:* `aggregations = [{ type = "count" }]`
- *Why:* `data_table` and `line_chart` already spell it plural, so it reads the same across widgets.
- *Does the API restrict this too?* No. Terraform-only rename, mapped back to `aggregation` on send.

**2. Enum values in lower case**
- *API (wire):* `"dataModeType": "DATA_MODE_TYPE_ARCHIVE"`
- *Terraform:* `data_mode_type = "archive"`
- *Why:* every other resource here does this, so users don't learn a different spelling per field.
- *Does the API restrict this too?* No. Terraform-only, mapped back on send.

**Empty lists are now rejected when you plan, not when you apply**
- *What the user saw:* `thresholds = []` passed the plan and then broke the **first** apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`. Reproduced against a real environment.
- *What changed:* it fails at plan, naming the attribute. Every list this widget exposes is guarded.
- *Why:* an empty list and an absent list are different values. We send empty, the API stores nothing, the refresh reads back null — and Terraform rejects the mismatch.
- *Does the API restrict this too?* Yes. The API has no way to hold an empty list, so there is nothing that could round-trip.

**Two deprecated fields kept on purpose: `interpretation` and `value_field`**
The API deprecates both — `interpretation` in favour of the visualization block, `value_field` in favour of `value_fields` — but dashboards out there still set them. Modelling them means those import cleanly and a later `apply` won't silently reset them. Both carry a deprecation notice, and deleting either from your config clears it rather than keeping the old value.

**`decimal_precision` accepts 0 to 15**
- *What it is:* how many digits to show after the decimal point.
- *API:* the proto documents `minimum: 0, maximum: 15`.
- *Terraform:* values outside that range are refused when you plan, naming the attribute.
- *Does the API restrict this too?* **No** — I applied `16` against a real environment and it was accepted and stored. This is a guard against the documented limit, not a mirror of backend behaviour, and the code says so where a future reader will see it.

## Scope

**Reused, not rewritten.** The query, filter and observation-field schemas are the same shared ones the existing widgets use.

**Two shared fixes, both the same bug as above.** `keypath` on observation fields and `filters` on spans queries accepted an empty list at plan and then failed the apply — reproduced against a real environment for both. This widget exposes both, so leaving them would ship a known bug in the new surface, and copying the schemas to fix them here only would be worse. **Not breaking, and no one-time diff on upgrade:** the only config they newly reject is one that already fails at apply today. No existing config, test or example sets either list to empty.

**The API-coverage manifest is reclassified.** The repo tracks every branch of the API and how the provider covers it. `dynamic`, its four query branches and `stat` move from "only usable through `content_json`, no import" to "covered by a test". The other visualizations stay as they are. Without this the coverage suite keeps passing while reporting the new surface as missing.

**Schema stays at version 4.** Adding an optional attribute needs no state upgrader — asserted, not assumed, below.

The remaining pre-existing bugs in shared enum maps and object validators are a separate PR, deliberately left out.

## Evidence

`TestAccCoralogixResourceDashboardDynamicStatWidget` runs against a real Coralogix environment on the repo's enforced create/update/import lifecycle: all four query sources, the metrics query with its optional fields both set and both left out, an update, a step that removes optional fields, and a verified import. Every step asserts the next plan is empty and reads the dashboard back from the API. **Every field of `stat` is set and checked**, so each one is proven to survive the trip out to the backend and back.

`TestAccCoralogixResourceDashboardDynamicRejectsEmptyLists` — `thresholds`, `category_fields`, `value_fields`, an observation field's `keypath` and a spans query's `filters` set to `[]` are each refused at plan.

`TestCurrentSchemaDecodesNarrowDynamicStatState` reads a stat widget stored before these fields existed, using the current schema — the evidence for keeping version 4. Honest limit: it can only fail for a type change or a removal, not for a purely additive one.

`TestFlattenDashboardRejectsDynamicWidgetWithLegacyTopLevelQuery` covers the deprecated-query case above. Worth stating how that one was found, because no test in this PR could have: a reviewer read the flatten and asked what happens to the old field. I then authored the legacy shape through `content_json` and read it back from a real environment, which returned the deprecated `query` populated and `query_definitions` empty — so the dashboard state is reachable, not theoretical. Without the guard the widget's only data source was dropped silently and a later apply would have erased it from the live dashboard. The test was checked by removing the guard and confirming it fails.

Unit tests in `dynamic_test.go` check `stat` and the query union survive expand → flatten. The example passes `terraform validate`; docs are regenerated.

**Checked in the product, not just the API.** The example was applied to a real account and the widget renders, with the threshold colours working — `14K` in red against the `1000` threshold. Two things that only a UI check could surface:

- On an account with no matching data the widget reports *"Data doesn't match this widget type"* rather than *"No data found"*. Nothing is wrong with the configuration; the message is just misleading when the account is empty.
- **Opening a `dynamic` `stat` in the UI editor rewrites it to a `stat_card`** and saves that. After that, this branch can no longer read the dashboard — the read fails with the diagnostic above, which is the fail-closed behaviour working as intended on a real payload. The follow-up PR adds `stat_card` and resolves it, so **the two should ship in the same release**; on its own this one leaves a dashboard that Terraform cannot read once anyone edits the widget in the product.

`TestDashboardProtoOneOfInventoryAgainstCheckout` cross-checks the manifest against the protobuf definitions. It needs a current sibling `cx-management-apis` checkout — with a stale one it stops at a count mismatch before reaching the manifest comparison, which is what happened to me at first. Against a current checkout it passes, so the reclassification above is verified against the protos and not only against the generated client.

## Not included, on purpose

- **The other visualizations** — `stat_card` next, then `table` and the time-series ones, the bar charts, and finally gauge, pie, hexagon, heatmap and geomap. Each brings its own tests, docs and example.
- **The deprecated single `query`** on the widget — use `query_definitions`, which holds the same queries and more (several, each named). Note it is *not* true that the API migrates the old field to the new one: I assumed that and it is wrong. A dashboard can hold a populated deprecated `query` with no `query_definitions` at all, so the read now refuses such a widget rather than dropping its data source — see the evidence section.
- **The shared-code bug fixes**, per the scope note above.

### DIFF
--- a/examples/resources/coralogix_dashboard/resource.tf
+++ b/examples/resources/coralogix_dashboard/resource.tf
@@ -894,6 +894,51 @@ resource "coralogix_dashboard" "widgets" {
           }
         }]
       }]
+      },
+      {
+        # A `dynamic` widget pairs one or more `query_definitions` with a single
+        # `visualization`. Each query picks exactly one source (logs here), and the
+        # widget picks exactly one visualization.
+        rows = [{
+          height = 19
+          widgets = [{
+            title = "dynamic stat - error volume"
+            definition = {
+              dynamic = {
+                query_definitions = [{
+                  name = "errors"
+                  query = {
+                    logs = {
+                      lucene_query = "coralogix.metadata.severity=\"5\" OR coralogix.metadata.severity=\"6\""
+                      group_by = [{
+                        keypath = ["subsystemname"]
+                        scope   = "label"
+                      }]
+                      aggregations = [{
+                        type = "count"
+                      }]
+                    }
+                  }
+                }]
+                time_frame = {
+                  relative = {
+                    duration = "seconds:900"
+                  }
+                }
+                visualization = {
+                  stat = {
+                    decimal_precision = 0
+                    threshold_type    = "absolute"
+                    thresholds = [
+                      { from = 0, color = "green" },
+                      { from = 1000, color = "red" },
+                    ]
+                  }
+                }
+              }
+            }
+          }]
+        }]
     }]
   }
 }
--- a/internal/provider/dashboards/dashboard_widgets/dynamic.go
+++ b/internal/provider/dashboards/dashboard_widgets/dynamic.go
@@ -866,6 +866,17 @@ func FlattenDynamic(ctx context.Context, dynamic *dashboardservice.WidgetsDynami
 		return nil, nil
 	}
 
+	// The deprecated top-level query has no typed equivalent, and the API stores
+	// and returns it verbatim rather than migrating it to query_definitions.
+	// Flattening such a widget would drop its only data source, so refuse the
+	// read instead of writing state that silently loses it.
+	if dynamic.Query != nil {
+		return nil, diag.Diagnostics{diag.NewErrorDiagnostic(
+			"Unsupported Dashboard Widget Definition",
+			"The dynamic widget uses the deprecated top-level `query`, which this provider cannot represent as typed HCL. Manage this dashboard with `content_json`, or move the widget to `query_definitions`, which holds the same queries.",
+		)}
+	}
+
 	queryDefinitions, diags := flattenDynamicQueryDefinitions(ctx, dynamic.GetQueryDefinitions())
 	if diags.HasError() {
 		return nil, diags
--- a/internal/provider/dashboards/dashboard_widgets/dynamic.go
+++ b/internal/provider/dashboards/dashboard_widgets/dynamic.go
@@ -17,7 +17,6 @@ package dashboard_widgets
 import (
 	"context"
 	"fmt"
-	"math"
 	"strings"
 
 	"github.com/coralogix/terraform-provider-coralogix/internal/utils"
@@ -334,8 +333,9 @@ func dynamicStatSchema() schema.Attribute {
 			"decimal_precision": schema.Int64Attribute{
 				Optional: true,
 				Validators: []validator.Int64{
-					int64validator.Between(0, math.MaxInt32),
+					int64validator.Between(0, 15),
 				},
+				MarkdownDescription: "How many digits to show after the decimal point. Valid values are 0 to 15.",
 			},
 			"display_series_name": schema.BoolAttribute{
 				Optional: true,
@@ -395,8 +395,10 @@ func dynamicStatSchema() schema.Attribute {
 			},
 			"unit": UnitSchema(),
 			"value_field": schema.SingleNestedAttribute{
-				Attributes: ObservationFieldSchema(),
-				Optional:   true,
+				Attributes:          ObservationFieldSchema(),
+				Optional:            true,
+				DeprecationMessage:  "Deprecated: superseded by value_fields.",
+				MarkdownDescription: "Deprecated: superseded by `value_fields`, which holds the same observation fields. Retained at full fidelity so dashboards that still set it can be imported and updated without losing it.",
 			},
 			"value_fields": schema.ListNestedAttribute{
 				Optional: true,
--- a/internal/provider/dashboards/dashboard_widgets/dynamic.go
+++ b/internal/provider/dashboards/dashboard_widgets/dynamic.go
@@ -1,0 +1,1183 @@
+// Copyright 2025 Coralogix Ltd.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dashboard_widgets
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"strings"
+
+	"github.com/coralogix/terraform-provider-coralogix/internal/utils"
+
+	dashboardservice "github.com/coralogix/coralogix-management-sdk/go/openapi/gen/dashboard_service"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/objectvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+var (
+	dashboardSchemaToProtoMetricsEditorMode = map[string]dashboardservice.MetricsQueryEditorMode{
+		utils.UNSPECIFIED: dashboardservice.METRICSQUERYEDITORMODE_METRICS_QUERY_EDITOR_MODE_UNSPECIFIED,
+		"builder":         dashboardservice.METRICSQUERYEDITORMODE_METRICS_QUERY_EDITOR_MODE_BUILDER,
+		"text":            dashboardservice.METRICSQUERYEDITORMODE_METRICS_QUERY_EDITOR_MODE_TEXT,
+	}
+	dashboardProtoToSchemaMetricsEditorMode = utils.ReverseMap(dashboardSchemaToProtoMetricsEditorMode)
+	dashboardValidMetricsEditorModes        = utils.GetKeys(dashboardSchemaToProtoMetricsEditorMode)
+
+	dashboardSchemaToProtoMetricsSeriesLimitType = map[string]dashboardservice.MetricsSeriesLimitType{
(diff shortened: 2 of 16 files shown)


dmenilo1:
@codex review

dmenilo1:
@codex review

dmenilo1:
@codex review

dmenilo1:
@codex review

examples/resources/coralogix_dashboard/resource.tf
```diff
@@ -894,6 +894,51 @@ resource "coralogix_dashboard" "widgets" {
           }
         }]
       }]
+      },
+      {
+        # A `dynamic` widget pairs one or more `query_definitions` with a single
+        # `visualization`. Each query picks exactly one source (logs here), and the
+        # widget picks exactly one visualization.
+        rows = [{
```
markshleifer-coralogix, line 902: Can you please check that this specific example works well in the UI as well? i.e:
Create it via TF, then open the dashboard in the UI and make sure it renders without errors.

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -0,0 +1,1183 @@
...
+				},
+			},
+			"custom_unit": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "Custom unit label. Takes effect only when `unit` is `custom`.",
+			},
+			"decimal_precision": schema.Int64Attribute{
+				Optional: true,
+				Validators: []validator.Int64{
+					int64validator.Between(0, math.MaxInt32),
```
markshleifer-coralogix: Can we use the same limit the proto has?
```
google.protobuf.Int32Value decimal_precision = 11 [(grpc.gateway.protoc_gen_openapiv3.options.openapiv3_field) = {
  description: "Number indicating the decimal precision of the numeric values, within range 0-15"
  example: "2"
  minimum: 0
  maximum: 15
}];
```

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -0,0 +1,1183 @@
...
+							Optional: true,
+						},
+						"label": schema.StringAttribute{
+							Optional: true,
+						},
+					},
+				},
+			},
+			"unit": UnitSchema(),
+			"value_field": schema.SingleNestedAttribute{
```
markshleifer-coralogix, line 397: `value_field` is deprecated at the proto level:
```
v1.common.ObservationField value_field = 1 [deprecated = true];
```
Should be treated the same way as `interpretation`.

internal/provider/dashboards/dashboard_widgets/dynamic.go
dmenilo1: Done — `Between(0, 15)` on `stat` here, and on `stat_card` in #645, with a description stating the range.

One thing worth recording: **the API does not enforce this.** I applied `decimal_precision = 16` against a real environment and it was accepted and stored, with no drift on refresh. So this is a plan-time guard against the documented limit rather than a mirror of backend rejection — there's a comment above the test saying so, in case someone later reads the validator as backend-enforced.

Not breaking: the dynamic widget hasn't shipped, so no configuration can be invalidated. `decimal_precision` is also exposed only by this widget — the legacy widgets don't surface it — so nothing else is affected.

Covered by `TestAccCoralogixResourceDashboardDynamicRejectsOutOfRangeDecimalPrecision` (`16` and `-1`).

Separately: my local `cx-management-apis` checkout was stale and had only the prose "within range 0-15", not the `minimum`/`maximum` keys you quoted. After updating it your annotations are there exactly as pasted. That also fixed `TestDashboardProtoOneOfInventoryAgainstCheckout`, which had been failing locally on a count mismatch — it stops there before reaching the manifest comparison, so with a current checkout that comparison now actually runs and the manifest reclassification in these PRs is verified against the protos, not just the generated client.

internal/provider/dashboards/dashboard_widgets/dynamic.go
dmenilo1: Done — same treatment as `interpretation`: a `DeprecationMessage` plus a description explaining it's superseded by `value_fields`, and that it's retained so dashboards which set it import and update without losing it.

I swept the proto for the rest of this class rather than just the field you flagged. In this PR's scope: `Dynamic.query` is deprecated and deliberately not modelled at all, `interpretation` already carried a notice, and `Stat.value_field` was the only gap. `StatCard` has no deprecated singular, so #645 needs nothing here. The remaining ones — `Gauge.value_field`, `Gauge.show_inner_arc` / `show_outer_arc`, and `Visualization.time_series_lines` — belong to visualizations that later PRs in this series add, and they're noted for those.

examples/resources/coralogix_dashboard/resource.tf
dmenilo1: Checked, and it was worth asking — it found something none of the automated tests could.

**The example renders correctly.** Applied to a real account, the widget shows `14K` in red, so the thresholds are working as configured.

The one thing only a UI check could surface: **opening the widget in the UI editor rewrites `stat` into `stat_card`** and saves that. After that, this branch can no longer read the dashboard — `terraform plan` fails with the unsupported-visualization diagnostic. That's the fail-closed behaviour working as designed on a real payload, but it means **#641 and #645 should ship in the same release**: on its own, this PR leaves a dashboard that Terraform cannot read as soon as anyone edits the widget in the product. I've added that to both descriptions.

Reproducible if you want to see it: create a `dynamic` stat through Terraform, open it in the UI, then run `terraform plan`.

markshleifer-coralogix: > After that, this branch can no longer read the dashboard — terraform plan fails with the unsupported-visualization diagnostic

Can you explain why it fails?

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -0,0 +1,1185 @@
...
+
+	return expanded, diags
+}
+
+func FlattenDynamic(ctx context.Context, dynamic *dashboardservice.WidgetsDynamic) (*WidgetDefinitionModel, diag.Diagnostics) {
+	if dynamic == nil {
+		return nil, nil
+	}
+
+	queryDefinitions, diags := flattenDynamicQueryDefinitions(ctx, dynamic.GetQueryDefinitions())
```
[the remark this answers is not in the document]
dmenilo1: Confirmed and fixed in f494eb2.

You were right that the flatten ignores it, and I checked whether the state is actually reachable rather than assuming. My PR description claimed `query_definitions` "is what the API stores" — that was wrong. Authoring the legacy shape through `content_json` and reading it back from a real environment returns:

```
Query (deprecated, top level) populated: true
QueryDefinitions count:                  0
```

So the API stores the deprecated field verbatim and does not migrate it. Two consequences, both worse than I'd assumed: the widget's only data source was dropped on read, and a later apply would have written it back without one, erasing the query from a live dashboard. The resulting state also violated the schema, since `query_definitions` is `Required` with `SizeAtLeast(1)`.

Took the diagnostic option rather than hydrating the deprecated field — it reuses what this function already returns for visualization variants it cannot represent, and `query_definitions` supersedes the field for anything new.

Covered by `TestFlattenDashboardRejectsDynamicWidgetWithLegacyTopLevelQuery` with a `content_json_dynamic_legacy_query.json` fixture. I verified the test fails with the guard removed (`flatten returned partial state`) rather than passing either way. `Dynamic.query` is the only deprecated query field in the API definition, and no existing fixture uses that shape, so nothing else changes. Description corrected.

examples/resources/coralogix_dashboard/resource.tf
dmenilo1: Every `plan` starts by **refreshing** — reading the real dashboard and converting it into state. The API describes a dynamic widget's visualization as a union: exactly one of `stat`, `stat_card`, `table`… is set. So the provider switches on which one it finds:

```go
switch {
case visualization.Stat != nil:  // convert it
default:                         // error: cannot represent this variant
}
```

This branch handles only `stat`. Once the UI has rewritten the widget, the API returns `statCard`, which hits `default`. Refresh returns an error instead of state and **the plan aborts before computing any diff**. `apply` and `destroy` refresh too, so all three fail.

**Why it errors instead of skipping the widget.** Deliberate. If the provider ignored the unrecognised visualization, it would build state describing a widget with *no* visualization — and Terraform would then "fix" that on the next apply, overwriting the `stat_card` the UI made. A clear error beats silent destruction.

**What #645 changes.** `stat_card` becomes a handled case, so refresh succeeds and you get an ordinary diff — config says `stat`, dashboard says `stat_card` — and you choose. A visible diff instead of a dead end. That is why both belong in the same release.

One correction to my note above: "can no longer read" understates it. With this PR alone the resource is **stranded** — `plan`, `apply` and `destroy` all fail. We hit this for real while destroying a test dashboard.

markshleifer-coralogix (approved):

PR #645, opened 2026-08-13
feat(dashboards): add the stat_card visualization to the dynamic widget (CX-50151)

dmenilo1:
## What this does

Adds the second visualization to the `dynamic` dashboard widget: `stat_card`. A stat card shows one headline number with its own title and label, each of which can read a field, use template text, or show the result of a colour mapping.

```hcl
visualization = {
  stat_card = {
    unit = "milliseconds"
    title         = { template_text = "p99 latency" }
    primary_value = { observation_field = { keypath = ["duration"], scope = "metadata" } }
    color_label_mapping = {
      range = {
        thresholds = [{ from = 0, color = "green" }, { from = 500, color = "red" }]
      }
    }
  }
}
```

Builds on #641, which added the widget and its first visualization. **Set this PR's base to `master` once #641 merges** — it targets that branch so the diff shows only the stat card.

## Where to look

| What | Where |
|---|---|
| The whole visualization | `dynamic_stat_card.go` (new) |
| Where it plugs in | `dynamic.go` — the five places every variant registers: the schema map, the attribute-type map, both dispatch switches, and the exactly-one-of list |
| Models | `common.go` |

**The thing worth checking:** the write direction (HCL → API) and the read direction (API → state) must carry the same fields. If the read drops a field the write sends, `import` loses it and the next `apply` sends it back empty, clearing the setting on the server.

**`mapped_values` is a `true`/absent switch, not data — and the generated code says otherwise**

- *Product:* the title or label can show the label produced by `color_label_mapping` instead of a value read from the data.
- *API:* the proto is `message MappedValues {}` — a marker with **no fields at all**. On the wire it is `{"mappedValues": {}}`: the empty object selects the option, it is not settings.
- *Terraform:* `mapped_values = true`
- *Why:* a boolean is the natural HCL for a marker, and this widget already does exactly this for `min_max.auto`, which is the same kind of empty marker.
- *Does the API restrict this too?* Yes. The API has nowhere to put values here, so nothing is lost.

This one is worth a reviewer's attention because the generated Go client types the field as `map[string]interface{}`, which reads like free-form JSON. An earlier draft took that at face value and modelled it as a JSON-object string, with a custom Terraform type to compare JSON semantically. The API rejects that outright:

```
400 Bad Request — proto: (line 1:2286): unknown field "a"
```

Reading the proto settled it. The custom type is **not** in this PR — a boolean replaces roughly a hundred lines of it plus its tests.

**Enum values in lower case** — `color = "green"`, `color_by = "value"`, `threshold_type = "absolute"`, where the API spells them `COLOR_SOLID_TYPE_GREEN` and so on. Terraform-only, mapped back on send; the same convention as every other resource here.

## Three API rules that used to fail at apply, now caught at plan

Each was found by applying against a real environment, and the quoted text is the API's own error.

**A colour mapping needs exactly one of `range`, `value`, `regex`.** Setting none gave `mapping type must be defined as range, value, or regex`; setting two gave `at most one of [range, value, regex]`. Both now name the attribute at plan time.

**An element cannot both read a field and show the mapping result.** `observation_field` together with `mapped_values` gave `at most one of [observationField, mappedValues]`. Same for display-name template variables. Setting neither is valid, so this rejects only the combination.

**`primary_value` can never use `mapped_values`,** because it is the value the mapping reads from: `primary value type cannot be mapped values (primary value is the source for mapping)`. The attribute stays in the schema for `title` and `label` and is refused on `primary_value` with that reason.

None of these three rules is expressible in the generated types — they are backend behaviour, so only a real apply finds them.

## Scope

**Empty lists are rejected, matching the rest of the widget.** Five lists this visualization exposes had no guard — `category_fields`, `value_fields`, `template_variables`, colour-mapping `sections` and `thresholds`. Without one, `category_fields = []` passes the plan and then fails the apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`, because an empty list and an absent list are different values and the API cannot store the former.

**One shared change: `thresholds` now has a single definition.** #641 declared the thresholds block inline for `stat`; the colour mapping needs the same block. Rather than a second copy that can drift, both now call one shared schema. Identical shape, and the guard above applies to both. No user-visible change.

**The API-coverage manifest is reclassified.** The repo tracks every branch of the API and how the provider covers it, and nothing detects staleness when the provider *gains* support — so it is done by hand. Ten branches move to "covered by a test": the stat card itself, both ways an element or a template variable gets its text, all three colour-mapping types, and both min/max modes. One explanation was also wrong before this PR: min/max was recorded as reachable only from geomap, but the colour mapping reaches it too.

**Schema stays at version 4.** Adding optional attributes needs no state upgrader — asserted, not assumed, below.

## Evidence

`TestAccCoralogixResourceDashboardDynamicStatCardWidget`, against a real Coralogix environment, on the repo's enforced create/update/import lifecycle: three stat cards covering all three colour-mapping types, both ways of filling an element, and template variables. Every step asserts the next plan is empty, that updates are updates rather than replacements, and **reads the dashboard back from the API** to check what the backend stored. Import runs the same backend check. One update switches the min/max from an explicit range to the automatic marker, so both arms of that union and the switch between them are exercised; a later step removes the optional enums and asserts they fall back rather than sticking.

**Three things a reviewer might expect to be restricted are deliberately not.** Each looked like a possible defect and each turned out fine when applied for real, so no validator was added — a validator that rejects a config the API accepts is the harder mistake to undo. The API accepts `custom_unit` alongside a non-custom `unit` and stores it (now asserted in the test, rather than left to a one-off check); a colour mapping whose `sections` list is omitted; and an empty template-variable entry. None produced an inconsistent-result error on apply.

`TestAccCoralogixResourceDashboardDynamicStatCardRejectsEmptyLists` covers the five lists. `...RequiresOneColorMappingType` and `...RejectsInvalidVisualElements` cover the three rules above, none and two arms included.

`TestCurrentSchemaDecodesNarrowDynamicStatCardState` reads a stat card stored before these attributes existed, using the current schema — the evidence for keeping version 4. Honest limit: it can only fail for a type change or a removal, not for a purely additive one.

Unit round-trips in `dynamic_test.go` cover the stat card and its automatic min/max. The data source inherits the schema from the resource, so it needs no separate wiring. The example passes `terraform validate`; docs are regenerated.

`TestDashboardProtoOneOfInventoryAgainstCheckout` cross-checks the manifest against the protobuf definitions, and needs a current sibling `cx-management-apis` checkout; with a stale one it stops at a count mismatch before reaching that comparison. Against a current checkout it passes, so the ten reclassified branches are verified against the protos, not only against the generated client.

**`decimal_precision` accepts 0 to 15**, matching the proto's documented `minimum: 0, maximum: 15` and the same change made to `stat` in #641. Worth knowing: the API accepts `16` — I applied it — so this guards a documented limit rather than mirroring backend rejection.

## Why this one matters more than "one more visualization"

Opening a `dynamic` `stat` widget in the Coralogix UI editor **rewrites it to a `stat_card`** and saves that. Verified by hand: create a stat through Terraform, open it in the product, and #641 alone can no longer read the dashboard — `terraform plan` fails with its unsupported-visualization diagnostic.

This PR is what closes that gap, so **the two should ship in the same release**. #641 on its own works right up until someone edits the widget in the UI.

## Not included, on purpose

- **A custom JSON string type** — see the `mapped_values` note above. It is not needed, so it is gone.
- **The remaining visualizations** — `table` and the time-series ones next, then the bar charts, then gauge, pie, hexagon, heatmap and geomap. Each brings its own tests, docs and example.
- **The shared-code bug fixes** in the remaining enum maps and object validators, which are their own PR.


internal/provider/dashboards/dashboard_widgets/dynamic_stat_card.go
```diff
@@ -0,0 +1,857 @@
...
+			"color_label_mapping": dynamicColorLabelMappingSchema(),
+			"custom_unit": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "A free-text unit label. Documented as taking effect only when `unit` is `custom`.",
+			},
+			"decimal_precision": schema.Int64Attribute{
+				Optional: true,
+				Validators: []validator.Int64{
+					int64validator.Between(0, 15),
+				},
```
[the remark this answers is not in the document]
dmenilo1: Deliberate. The `16` note is mine, from probing a real environment — the API accepts it, but `0-15` is the proto's documented contract (`minimum: 0, maximum: 15`) that the backend just doesn't police. A value above it isn't a valid config we're blocking; it's out of range and would render incorrectly, so rejecting it at plan time names the attribute rather than producing a confusing widget.

Your narrow case is real: a card created via API or `content_json` with `> 15` can be imported but not expressed in HCL. That needs hand-authored out-of-range input — the UI won't produce it. If it ever surfaces, relaxing the upper bound is the fix, and this thread is the record of why it was there.

Requested by the API owner in the `decimal_precision` thread on #641.

dmenilo1:
@codex review

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@
 #### resource/coralogix_dashboard
 - FEAT: Add typed HCL support for the `dynamic` widget definition: `query_definitions` with the logs/spans/metrics/data_prime query union, a top-level `time_frame` and `interpretation`, and the `stat` visualization at full fidelity. The remaining visualizations follow in later releases; a dynamic widget using one of them, or using the deprecated top-level `query` instead of `query_definitions`, fails the read with a clear diagnostic instead of writing state that silently drops it. Previously every `dynamic` widget failed import and data-source reads.
 - DEPRECATION: Mark `dynamic.visualization.stat.value_field` as deprecated, matching the API. Use `value_fields`; the singular form is still read and written so existing dashboards import without losing it.
+- FEAT: Add the `stat_card` visualization to the `dynamic` widget: the `title`, `label` and `primary_value` elements (each reading a field or using template text), display-name template variables, and `color_label_mapping` with its `range`, `value` and `regex` variants.
```
markshleifer-coralogix: I don't think we need a separate change-log for each PR. Probably one changelog for the whole `dynamic` widget is enough.

internal/provider/dashboards/dashboard_widgets/dynamic_stat_card.go
```diff
@@ -0,0 +1,857 @@
...
+				Computed: true,
+				Default:  stringdefault.StaticString(utils.UNSPECIFIED),
+				Validators: []validator.String{
+					stringvalidator.OneOf(DashboardValidLegendBys...),
+				},
+				MarkdownDescription: fmt.Sprintf("How the legend is grouped. Valid values are: %s.", strings.Join(DashboardValidLegendBys, ", ")),
+			},
+			"primary_value": dynamicStatVisualElementSchema(false),
+			"title":         dynamicStatVisualElementSchema(true),
+			"unit":          UnitSchema(),
```
markshleifer-coralogix, line 81: `UnitSchema()` uses `DashboardSchemaToProtoUnit`, which is missing two values that are already in the SDK `CommonUnit` enum: `UNIT_PERCENT` and `UNIT_DATETIME_ISO`.
Gauge already maps `percent`.
A UI-saved stat card with either unit flattens to null, then the default writes `unspecified` on the next apply. This PR is the UI rewrite path from `stat` to `stat_card`, so those two keys should be added to the shared map (no SDK bump).

dmenilo1: Confirmed and fixed. The SDK accepts 22 `CommonUnit` values and the shared map had 20 — `UNIT_PERCENT` and `UNIT_DATETIME_ISO` were both absent — and your chain is exactly right: `flattenOptionalEnum` returns null for a value it cannot map, and the attribute's default then writes `unspecified` on the next apply.

I added a coverage test over the SDK's allowed values rather than just the two keys, and it found a third gap you didn't mention: `DashboardSchemaToProtoGaugeUnit` was missing `datetime_iso`. Both maps are now asserted complete, so an SDK bump that adds a unit fails the test instead of silently dropping reads of it.

No SDK bump needed, as you said. Scope note: this affects every widget type exposing a `unit`, not only `dynamic` — included here for the reason you gave, that this branch is the path by which the UI rewrites a stat into a stat card, so faithfully reading a UI-saved unit is exactly its job.

CHANGELOG.md
dmenilo1: Agreed, done. The seven per-PR bullets are now a single entry describing the dynamic widget as a whole — the query union, the time frame, `stat` and `stat_card`, the fail-closed reads, and the configurations now rejected at plan time instead of at apply.

Kept separate: the `value_field` deprecation and the unit-map fix, since neither is part of the feature description and the unit fix reaches widgets beyond `dynamic`.

#648 (the time-series visualizations) will pick up the same shape when it rebases.

markshleifer-coralogix (dismissed):

markshleifer-coralogix (approved):

PR #648, opened 2026-08-13
feat(dashboards): add the time-series visualizations to the dynamic widget (CX-50151)

dmenilo1:
## What this does

Adds the time-series visualizations to the `dynamic` dashboard widget: `time_series_lines_multi`, `time_series_bars`, and the deprecated singular `time_series_lines`.

```hcl
visualization = {
  time_series_lines_multi = {
    stacked_line       = "absolute"
    x_axis_time_format = "hh_mm"
    query_display_settings = [{
      query_id          = "9d1b7a4e-0000-4000-8000-00000000ab01"  # a query_definitions id
      scale_type        = "linear"
      unit              = "seconds"
      decimal_precision = 2
    }]
  }
}
```

Third in the series after the widget itself (#641) and `stat_card` (#645), both merged. `table` is up in #662; the bar, gauge, pie, hexagon, heatmap and geomap visualizations follow.

## Where to look

| What | Where |
|---|---|
| The three visualizations | `dynamic_time_series.go` (new) |
| The one behaviour change to existing schema | `dynamic.go`, `query_definitions[*].id` |
| Models | `common.go` |

## Why the deprecated variant is here, and not skipped

`time_series_lines` is deprecated in the API in favour of `time_series_lines_multi`. Modelling it is not only import fidelity — **opening the widget in the Coralogix UI rewrites the singular form into the multi form** and saves that.

Checked by hand: after editing in the UI, `terraform plan` returned a normal diff naming the swap, at exit 0. A release carrying only the replacement would instead fail the read on any dashboard whose widget someone had opened, which is what happens today with `stat`. Both variants shipping together is what avoids that.

Worth being precise about the reason, because "deprecated variant gets migrated to its replacement" is the wrong mental model. The UI picks whichever visualization matches the **shape of the data**, so the direction depends on the data rather than on which variant the API deprecates. Same behaviour, opposite direction, measured on the bar charts: a single-query result there is rewritten from `*_multi` **to** the singular form. So the rule is that both halves of a pair must ship together whichever way the arrow happens to point — not that deprecated variants get upgraded.

## The gap this closes

**`query_display_settings` could not be written at all.** Each entry styles one query and must name it through a required `query_id`, which has to match a `query_definitions[*].id`. That `id` was **`Computed` only** — the user was not allowed to set it, and its value did not exist until after apply. So there was never an id to reference: only a dashboard imported from the backend had both sides populated.

`query_definitions[*].id` is now **`Optional` and `Computed`** — still generated when omitted, so nothing changes for anyone not using this. The API accepts client-supplied query ids; its own `content_json` payloads carry them.

The acceptance test covers two queries on one chart with different display settings, and asserts both survive the round trip by reading the dashboard back from the API. What I verified in the UI is that the chart renders and the settings are stored against the right query ids; I did not do a pixel-level check that each line is styled differently.

**`y_axis_min` / `y_axis_max` use a semantic-equality number type**
- *Product:* the fixed ends of the chart's value axis.
- *API:* `float32`, not `float64` — `"yAxisMax": 99.5`.
- *Terraform:* `y_axis_max = 99.5`
- *Why:* a `float64` attribute round-tripping through `float32` can come back with a different decimal expansion and produce a permanent diff. A custom type compares them by value instead.
- *Does the API restrict this too?* It is the API's own precision; Terraform only avoids being tripped by it. The acceptance test covers a whole number and a fractional one.

**`series_count_limit` is a number in HCL, a string on the wire** — the API carries it as `"seriesCountLimit": "10"`. Terraform-only, converted on send.

**Empty lists are rejected when you plan, not when you apply.** Seven lists this surface exposes — `category_fields` and `value_fields` on lines, bars and each per-query settings entry, plus `query_display_settings` itself — had no guard, so `category_fields = []` passed the plan and then failed the apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`. An empty list and an absent one are different values, and the API cannot store the former.

## Scope

**One change to schema that already shipped:** `query_definitions[*].id` gains `Optional`. Additive, so no configuration breaks, and it is what makes `query_display_settings` usable at all.

**The API-coverage manifest is reclassified.** Three `Visualization` branches move from "content_json-only, no import" to "covered by a test". Nothing detects staleness when the provider *gains* support, so this is done by hand.

**Schema stays at version 4.** Adding optional attributes needs no state upgrader — asserted below.

**No unions in this surface.** None of these five API types carries an `at most one of` guard, so the oneof-arm default class that caused most of the trouble earlier in this series does not apply here.

## Evidence

`TestAccCoralogixResourceDashboardDynamicTimeSeriesWidgets`, against a real Coralogix environment, on the repo's enforced create/update/import lifecycle: all three visualizations, per-query display settings referencing an explicit query id, a whole and a fractional axis bound, an update, a step that removes the optional enums and asserts they fall back, and a verified import. Every step asserts the next plan is empty and reads the dashboard back from the API.

`...RejectsEmptyLists` covers all seven lists, one subtest each. `TestDataSourceSchemaPreservesResourceCustomTypes` covers the data-source fix by comparing every attribute type against the resource's. `TestCurrentSchemaDecodesNarrowDynamicTimeSeriesState` reads a time-series widget stored before these attributes existed, against the current schema — the evidence for keeping version 4. Honest limit: it can only fail for a type change or a removal, not for a purely additive one.

Every field of all five affected API models is modelled in both directions, so nothing is dropped on read and re-sent empty on the next apply.

**Checked in the product, not just the API**, and this is the part worth reading before you apply the example.

**All three visualizations display "Data doesn't match this widget type" until the widget is opened in the editor.** Opening it once makes the data appear, and it persists. This is existing product behaviour, not something this PR introduces or can fix: the UI runs a service that switches a dynamic widget's visualization to match the shape of the data when the widget enters edit mode. It is filed as BUGV2-6905 and closed as intended behaviour. The `table` visualization in #662 renders immediately, so this is specific to these three rather than universal to `dynamic`.

Three further things only a UI check surfaced:

- Time-series charts need `promql_query_type = "range"`. With the default the query returns a single point and the chart renders empty while the configuration looks entirely correct — I hit this myself. Now documented on the attribute and in the example.
- Editing a dynamic widget in the UI fills in optional settings the configuration does not set, so a later plan shows those being reset: `temporal_field`, `value_fields`, `allow_abbreviation`, `color_scheme`, `hash_colors`, `use_data_time_range` and `tooltip`. Ordinary Terraform semantics — the configuration is authoritative — but worth knowing; noted on the visualization.
- The UI also sets `query_definitions[*].query.metrics.editor_mode` to `text`, which a configuration that omits it resets to `unspecified` on the next apply. Pre-existing rather than introduced here, but these visualizations are what make it visible, since they are the ones people open in the editor.

## Not included, on purpose

- **`table`** — its own PR, #662, rather than doubling this one.
- **The bar, gauge, pie, hexagon, heatmap and geomap visualizations** — each with its own tests, docs and example.
- **The shared-code fixes**, which are a separate PR.


dmenilo1:
@codex review

dmenilo1:
@codex review

markshleifer-coralogix (approved):

PR #662, opened 2026-08-24
feat(dashboards): add the table visualization to the dynamic widget (CX-50151)

dmenilo1:
## What this does

Adds the `table` visualization to the `dynamic` dashboard widget. A table lists raw rows; `rules` restyle individual columns, each rule picking columns one way and applying one formatting change.

```hcl
visualization = {
  table = {
    columns  = [{ field = { keypath = ["applicationname"], scope = "label" } }]
    settings = { row_style = "one_line" }
    rules = [{
      name       = "bytes columns"
      rule_scope = { field_type = "number" }
      properties = [{
        definition = { units = { unit = "bytes", decimal_precision = 2 } }
      }]
    }]
  }
}
```

Targets `master`. #645 is merged, so the diff shows only the table.

## Where to look

| What | Where |
|---|---|
| The whole visualization | `dynamic_table.go` (new) |
| Where it plugs in | `dynamic.go` — the five places every variant registers: the schema map, the attribute-type map, both dispatch switches, and the exactly-one-of list |
| Models | `common.go` |

**The thing worth checking:** `rule_scope` and `definition` are both unions, and the two directions are enforced in different places. The plan-time validator is the only thing stopping an invalid combination, and the round-trip is the only thing proving no field is dropped. Neither can catch the other's bug — see below.

**A rule picks columns one way, and changes one thing — enforced at plan time**

- *Product:* a rule targets columns by name, by pattern, or by type, and applies a single formatting change to them.
- *API:* setting two alternatives returns `at most one of [...]`. Setting **none** is accepted — I applied a rule with no `ruleScope` and a property with no `definition` through `content_json`, and the backend stored both verbatim with an empty follow-up plan.
- *Terraform:* `ExactlyOneOfChildren` on both, naming the attribute during plan. It rejects two arms, and also rejects a block that is present with no arm — which the API would have accepted.
- *Why:* the API's error for two arms names generated JSON fields, which do not correspond to any path a user wrote. For the zero-arm case the validator is stricter than the backend: a rule that selects nothing does nothing, and catching it at plan time is more useful than storing it.
- *Does the API restrict this too?* Only for two arms. **Omitting the block entirely stays legal** and is how an existing dashboard with no scope is expressed — which is why neither container is `Required`.

Worth a reviewer's attention: **this rule exists nowhere in the generated types.** They give `RuleScope` three independent pointers rather than a discriminated union, so the write direction copies all three arms and the read direction returns all three. A round-trip test therefore passes whether or not the validator exists, and the validator passes whether or not the write direction is correct. The two are checked separately and deliberately — by `…RejectsMultipleUnionArms` and by the unit round-trip respectively.

**The read direction must not produce a union it would then reject.** The API also stores `ruleScope: {}` and `definition: {}` — present, with no alternative selected. Read back literally, those become objects the validator above refuses, so the dashboard could be imported and then never planned. Both now read back as absent. Same reasoning for table columns: the API stores a column field whose `keypath` is empty or missing, and normalises a missing one to an empty array, so `keypath` is optional here rather than reusing the shared field schema, which requires at least one segment. Without that, a dashboard this repo already applies in its own `content_json` tests could not be expressed as typed HCL.

The one visible consequence: a dashboard stored with `ruleScope: {}` has it normalised away on the first apply through typed HCL. The rendered table is unchanged — an unselected scope carries no behaviour — but the stored JSON differs from what was there before. An empty `keypath` needs no such warning: omitting it in HCL and having the API store `[]` are the same thing to the backend, and the pair round-trips with an empty plan.

**Enum values in lower case** — `row_style = "one_line"`, `alignment = "center"`, `unit = "bytes"`, where the API spells them `ROW_STYLE_ONE_LINE` and so on. Terraform-only, mapped back on send; the same convention as every other resource here.

**`row_style` falls back to `unspecified` when removed,** like every other optional enum in this package. Safe here because it is a plain field: the same default on a *union arm* is what makes the API reject a request as having more than one alternative set, which is why no arm in this PR carries one.

## Scope

**Empty lists are rejected, matching the rest of the widget.** Every list this visualization exposes — `columns`, `rules`, `properties`, `column_widths`, link `actions`, values-mapping `mappings` and threshold `values`. Without a guard, `columns = []` passes the plan and then fails the apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`, because an empty list and an absent list are different values and flatten returns the absent one. Rejecting at plan time is the smaller fix; making empty lists round-trip would change what the API is sent.

**`thresholds` is reused, not re-declared.** #645 made the thresholds block a single shared schema; the table's threshold property calls the same one. No user-visible change, and one fewer copy to drift.

**One test changed meaning rather than being added.** `TestFlattenDynamicRejectsUnmodeledVisualization` asserts that a visualization this provider version cannot model fails the read instead of writing partial state — so it has to point at something still unmodelled. It pointed at `table`; it now points at `gauge`, and the new `content_json_dynamic_unsupported_visualization.json` fixture is gauge-based for the same reason. The next visualization to land will need to move it again; that is the test working, not rotting.

**The API-coverage manifest is reclassified.** The repo tracks every branch of the API and how the provider covers it, and nothing detects staleness when the provider *gains* support, so it is done by hand. Twelve branches move to "covered by a test": the eight property definitions, the three rule scopes, and the table itself.

## Evidence

`TestAccCoralogixResourceDashboardDynamicTableWidget`, against a real Coralogix environment, on the repo's enforced create/update/import lifecycle. One rule per property-definition alternative, so every arm is exercised by a real apply. Every step **reads the dashboard back from the API** rather than trusting state, import included, and asserts the next plan is empty and that updates are updates rather than replacements. A later step removes the optional enums and asserts they fall back to `unspecified` instead of keeping the previous value.

`TestDynamicWidgetTableFullFidelityRoundTrip` covers write → read across every field, all eight definition arms and all three rule scopes, one arm per property so the shape matches what the API accepts. It runs in PR CI without credentials. I checked it actually fails when a field is dropped from the read direction, rather than passing by construction.

`…RejectsEmptyLists` covers all seven lists; `…RejectsMultipleUnionArms` covers two arms and a present-but-empty block, on both unions. `TestDynamicWidgetTableReadsShapesTheSchemaMustExpress` covers the three backend shapes above, and `TestAccCoralogixResourceDashboardDynamicTableColumnWithoutKeypath` applies a keypath-less column for real. I checked the two union cases fail without the fix; the keypath one cannot be caught by a read test, so `TestDynamicTableColumnKeypathIsOptional` asserts the schema directly.

All three acceptance tests pass against EU2. Unit tests and every dashboard acceptance shard are green on this PR.

**Verified in the product.** Applied against a real account, then opened in the Coralogix UI: the table renders on first dashboard load, and opening the widget in the editor leaves "Save changes" greyed out — the UI does not rewrite what the provider wrote. This is the first visualization in this series where that is true. `stat` is silently rewritten to `stat_card` on edit, which is why those two had to ship together; `table` has no equivalent, so it carries no release-pairing constraint.

**`decimal_precision` accepts 0 to 15,** matching the documented range on the units field and the same bound `stat` and `stat_card` already use. It was unbounded in this PR's first push, which was an inconsistency rather than a deliberate choice — the API itself accepts a larger value, so this guards a documented limit rather than mirroring a backend rejection.

**Two things a reviewer might expect to be restricted are deliberately not.** The API accepts `custom_unit` alongside a non-custom `unit`, and a rule whose `properties` list is omitted. Both looked like possible defects and both were fine when applied for real, so no validator was added — one that rejects a configuration the API accepts is the harder mistake to undo.

## Not included, on purpose

- **The remaining visualizations** — the bar charts, then gauge, pie, hexagon, heatmap and geomap. Each brings its own tests, docs and example.
- **Anything about the UI's auto-switching behaviour.** Worth knowing, because it bounds what this series can promise: the editor picks a visualization to match the shape of the data, so a widget can be moved to one this provider version does not model, after which the read fails by design. That is existing product behaviour, already filed and closed as intended, and not something this PR changes.


dmenilo1:
@codex review

dmenilo1:
@codex review

dmenilo1:
@codex review

markshleifer-coralogix (dismissed):

markshleifer-coralogix (approved):

PR #665, opened 2026-08-24
feat(dashboards): add the bar visualizations to the dynamic widget (CX-50151)

dmenilo1:
## What this does

Adds the four bar visualizations to the `dynamic` dashboard widget: `vertical_bars`, `vertical_bars_multi`, `horizontal_bars` and `horizontal_bars_multi`.

```hcl
visualization = {
  vertical_bars_multi = {
    colors_by  = "query"
    scale_type = "linear"
    sort_order = {
      order_direction = "desc"
      strategy        = { query_value = { query_id = "…" } }
    }
    query_field_settings = [{
      query_id    = "…"                                              # a query_definitions id
      value_field = { keypath = ["count_0"], scope = "user_data" }
    }]
    category_fields = [{ keypath = ["applicationname"], scope = "user_data" }]
  }
}
```

Stacked on #648, which this needs for its `Float32Type` — the bars axis bounds are `float32` on the wire, same as the time-series ones. **Set the base to `master` once #648 merges.**

## Where to look

| What | Where |
|---|---|
| The four visualizations | `dynamic_bars.go` (new) |
| Models | `common.go` |
| Shared helpers pulled across | `common.go`, `schema.go` — `ExpandColorsBy`, `FlattenColorsBy`, `ColorsBySchema` |

## Two fixes to the code this was extracted from

**`sort_order.strategy.category` was modelled as a JSON object string, and the API rejects that.** The generated client types it as `map[string]interface{}`, which reads like free-form JSON. The proto says otherwise:

```proto
message SortByCategory {
  option (...) = {json_schema: {description: "Sort by category."}};
}
```

An empty marker with no fields — the same shape as `mapped_values`, where this exact misreading was already made and corrected once. So `category` is now a `true`/absent bool using the same helper the stat card uses, and the custom JSON string type it depended on is gone. Nothing could have selected this strategy before: a JSON value comes back as `unknown field`.

**Every list the surface exposes now rejects an explicit empty value at plan time.** Seven of them, none previously guarded. Without one, `category_fields = []` passes the plan and fails the apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`, because an empty list and an absent list are different values and the API cannot store the former.

**A sort strategy picks one key, and the marker arm carries no settings**
- *Product:* order the bars either by their category or by the values of a chosen query.
- *API:* `SortStrategy` is a oneof of `SortByCategory` (empty) and `SortByQueryValue` (a query id), with a `strategy_type` discriminator the backend derives.
- *Terraform:* `strategy = { category = true }` or `strategy = { query_value = { query_id = … } }`
- *Why:* a bool is the natural HCL for a marker, and this widget already does exactly this for `mapped_values` and `min_max.auto`.
- *Does the API restrict this too?* Yes — `ExactlyOneOfChildren` mirrors the oneof. `strategy_type` stays read-only, since the backend sets it from whichever arm is present.

**`colors_by` is a single enum in HCL, a oneof of empty markers on the wire** — `colors_by = "stack"` becomes `{"stack": {}}`. Terraform-only; the same mapping the legacy bar-chart widgets already use, and their tests already cover all five arms.

## Scope

**One shared change beyond the new file:** the two visualization dispatch switches in `dynamic.go` are split into per-family helpers. With nine arms each they sat at cyclomatic complexity **20** against the pre-commit limit of 15 — which no CI job runs, so it only bites a maintainer with the hook installed. Each family is now its own small switch, and the remaining visualizations add a helper rather than growing one function. Behaviour is unchanged; the fail-closed read for an unmodelled visualization is still there.

**The API-coverage manifest is reclassified.** Four `Visualization` branches move to "covered by a test", and `SortStrategy` moves out of "API-only" entirely — both of its arms are now authorable. Nothing detects staleness when the provider *gains* support, so this is done by hand.

## Evidence

`TestAccCoralogixResourceDashboardDynamicBarsWidgets`, against a real environment, on the repo's enforced create/update/import lifecycle: all four visualizations, both sort-strategy arms, per-query field settings, a whole and a fractional axis bound, and a step that removes the optional enums and asserts they fall back rather than sticking. Every step reads the dashboard back from the API, import included, and the backend check asserts the sort strategy stored **exactly one** arm rather than trusting Terraform state.

`TestDynamicWidgetBarsFullFidelityRoundTrip` covers write → read across every field of all four, one subtest each, and runs in PR CI without credentials. I checked it fails when the `category` marker or the `float32` axis path is dropped, rather than passing by construction — which is how I caught the dispatch refactor silently dropping the `horizontal_bars_multi` arm from both directions.

`…RejectsEmptyLists` covers all seven lists; `…RejectsInvalidSortStrategy` covers both arms set and neither set; `…RejectsFalseCategoryMarker` covers `category = false`, which selects nothing — the union validator counts any non-null child as chosen while the write direction drops a false marker, so it needs the must-be-true validator rather than the union check alone.

## Verified in the product

**The editor rewrites bars between the singular and multi forms, by data shape.** A `*_multi` widget with a single query comes back as the singular form; with two queries it stays multi. This is the same UI behaviour as `stat` → `stat_card` and `time_series_lines` → `time_series_lines_multi`, but the direction here is the opposite, which is worth knowing because "the deprecated variant gets migrated to its replacement" is the wrong mental model — the UI picks whatever matches the data. Shipping all four together is what keeps the read working either way; a release with only some of them fails `plan`, `apply` and `destroy` on any widget someone has opened.

**Bar widgets display "Data doesn't match this widget type" until opened in the editor once.** Existing product behaviour, filed as BUGV2-6905 and closed as intended: the UI switches the visualization to match the data when the widget enters edit mode. Not introduced or fixable here.

**The example was wrong until the product corrected it.** A bar chart needs a value to plot, so the query has to aggregate — and the result column is named after the aggregation *and its index* (`count_0`, not `count`), while a grouped field arrives as a data column (`user_data`, not `label`). The example now uses both, with the reasons inline. Worth stating plainly: the acceptance tests passed against the API the whole time, because the API accepts these shapes happily and only the UI knows they plot nothing.

**Known limitation, not fixable in schema:** a `*_multi` configuration with one query will be rewritten to the singular form on the first UI visit. Use more than one query to hold a multi chart.

**Expect drift after a UI edit:** the editor fills in optional settings the configuration omits — legend, `allow_abbreviation`, `color_scheme`, `decimal_precision`, `hash_colors`, `max_bars_per_chart` — and adds a default `sort_order`. Ordinary Terraform semantics, since the configuration is authoritative.

## Not included, on purpose

- **The remaining visualizations** — gauge, pie, hexagon, heatmap and geomap. Each brings its own tests, docs and example.
- **The `Default: UNSPECIFIED` question.** Around 119 attributes across `internal/provider/dashboards/` reset to `unspecified` when omitted, so anything the UI writes flips back on the next apply. Pre-existing and repo-wide; its own PR, after working out which of them the UI actually writes.


dmenilo1:
@codex review

dmenilo1:
@codex review

markshleifer-coralogix (dismissed):

markshleifer-coralogix (approved):

PR #666, opened 2026-08-24
feat(dashboards): add the gauge and pie chart visualizations to the dynamic widget (CX-50151)

dmenilo1:
## What this does

Adds the `gauge` and `pie_chart` visualizations to the `dynamic` dashboard widget.

```hcl
visualization = {
  gauge = {
    unit           = "percent"
    min            = 0
    max            = 1000
    threshold_type = "absolute"
    arc_display    = { threshold_arc = true }
    thresholds = [
      { from = 0, color = "green", label = "ok" },
      { from = 500, color = "red", label = "high" },
    ]
    value_field = { keypath = ["count_0"], scope = "user_data" }
  }
}
```

Branched from `master` and independent of #648, #662 and #665 — it can be reviewed and merged in any order relative to them.

## Where to look

| What | Where |
|---|---|
| Both visualizations | `dynamic_gauge_pie.go` (new) |
| Models | `common.go` |

**The easiest surface in this series so far:** neither `Gauge` nor `PieChart` carries a oneof in the proto, so none of the union and empty-marker handling that produced most of the bugs in the bar and stat-card visualizations applies here.

**The pie chart's label source needs its own enum map, despite one already existing**
- *Product:* where the slice label sits — inside the slice, or on the stack.
- *API:* the SDK has **two distinct types** with identical values: `WidgetsPieChartLabelSource` for the legacy pie-chart widget, and `VisualizationPieChartLabelSource` for the dynamic widget's.
- *Terraform:* `label_source = "inner"`
- *Why:* the exported `DashboardSchemaToProtoPieChartLabelSource` is typed to the legacy enum and will not compile against the dynamic one. The new local map is not duplication, and the reason is recorded next to it so nobody consolidates them later.
- *Does the API restrict this too?* Yes, both are closed enums; the schema mirrors them.

**`color_scheme` is documented but not restricted.** The API types it as a free-form string and stores an unrecognised value verbatim — I applied `"not-a-real-scheme"` and the follow-up plan was clean. `DashboardValidColorSchemes` is a hardcoded provider-side list rather than an SDK enum, so a `OneOf` validator would reject configurations the backend accepts and would need editing whenever a scheme is added. The attribute names the schemes the product ships instead, which gives discoverability without a false ceiling. Worth knowing: a typo applies no scheme and nothing warns.

**`decimal_precision` accepts 0 to 15,** matching the proto and the bound `stat` and `stat_card` already use. It was unbounded in this PR's first push. Note the API itself accepts `16` — I applied one — so this guards a documented limit rather than mirroring a backend rejection.

## Scope

**Empty lists are rejected at plan time, matching the rest of the widget.** Four of the five lists this surface exposes had no guard; `thresholds` already inherited one from the shared schema #645 introduced. Without a guard, `category_fields = []` passes the plan and fails the apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`, because an empty list and an absent list are different values and the API cannot store the former.

**All five `repeated` fields in both protos are modelled** — gauge's `value_fields`, `category_fields` and `thresholds`, and the pie chart's `category_fields` and `sub_category_fields` — so nothing is dropped on read and re-sent empty on the next apply.

**The API-coverage manifest is reclassified.** Two `Visualization` branches move to "covered by a test". Nothing detects staleness when the provider *gains* support, so this is done by hand.

**One test carried over from #665:** `dynamic_list_guard_test.go`, which walks the schema and requires a minimum-size validator on every list attribute. It is here because this branch is cut from `master` and would otherwise not have it — it is what makes the four guards above verifiable rather than asserted. If #665 merges first, expect an identical-content conflict on that one file.

## Evidence

`TestAccCoralogixResourceDashboardDynamicGaugeAndPieWidgets`, against a real environment, on the repo's enforced create/update/import lifecycle: both visualizations, the gauge's arc display and thresholds, the pie chart's label definition and slice limits, and a step that removes the optional enums and asserts they fall back to `unspecified` rather than sticking. Every step reads the dashboard back from the API, import included.

`TestDynamicWidgetGaugeAndPieFullFidelityRoundTrip` covers write → read across every field of both, and runs in PR CI without credentials. I checked it fails when `show_inner_arc` or `label_source` is dropped from the read direction, rather than passing by construction.

`…RejectsEmptyLists` covers all five lists, one subtest each.

## Verified in the product

Applied against a real account and opened in the UI.

**Neither visualization is rewritten by the editor.** `gauge` stayed `gauge` and `pie_chart` stayed `pieChart` after opening and saving both. That is the expected outcome rather than a lucky one: the UI switches a dynamic widget's visualization to match the shape of the data, and unlike `stat`, the time-series pair and the bar pair, these two have no sibling variant to switch to. **So they carry no release-pairing constraint and can ship alone.**

**The pie chart renders on first load; the gauge needs one editor visit.** Existing product behaviour, filed as BUGV2-6905 and closed as intended — the widget displays "Data doesn't match this widget type" until it enters edit mode once. Not introduced or fixable here. Across this series the split is now: `table` and `pie_chart` render immediately; `stat`, the time-series, the bars and `gauge` need a visit.

**Both need an aggregating query.** A gauge plots one value and a pie chart needs slice values, so the query has to aggregate — and the result column is named after the aggregation *and its index* (`count_0`, not `count`), while a grouped field arrives as a data column (`user_data`, not `label`). The example uses both. This is worth stating because the acceptance tests pass against the API either way: the API accepts a configuration that plots nothing.

**Expect drift after a UI edit:** the editor fills in optional settings the configuration omits — legend, `allow_abbreviation`, `decimal_precision`, `color_scheme`, `value_arc`, `show_percentage`, `max_slices_per_stack`. Ordinary Terraform semantics, since the configuration is authoritative.

## Not included, on purpose

- **`hexagon_bins`, `heatmap` and `geomap`** — one more PR. They share a file upstream, and heatmap needs the `float32` semantic-equality type from #648, so that PR follows rather than being split around a dependency that resolves itself.
- **The `min_slice_percentage` and `max_slices_per_*` bounds.** The proto documents `min_slice_percentage` as 0–100 and a minimum of 1 for the slice and bar caps, while the schema allows the full int32 range. A separate pass, since it is a different attribute set from the one corrected here and applies across visualizations already merged.
- **The `Default: UNSPECIFIED` question.** Around 119 attributes across `internal/provider/dashboards/` reset to `unspecified` when omitted, so anything the UI writes flips back on the next apply. Pre-existing and repo-wide; its own PR, after working out which of them the UI actually writes.


dmenilo1:
@codex review

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,552 @@
...
+				},
+			},
+			"custom_unit": schema.StringAttribute{
+				Optional: true,
+			},
+			"decimal_precision": schema.Int64Attribute{
+				Optional: true,
+				Validators: []validator.Int64{
+					int64validator.Between(0, math.MaxInt32),
+				},
```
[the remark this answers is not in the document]
dmenilo1: Fixed in 7a75afd — both are now `Between(0, 15)` with a matching description.

Worth adding: every one of the 15 `decimal_precision` fields in `dynamic.proto` documents `maximum: 15`, so this was a consistency gap in six places rather than two. The same defect was sitting unflagged in the bar visualizations, fixed on that branch in `d1c30ab`.

The `max_bars_per_chart` and `max_slices_per_bar` validators in that file are deliberately left at the int32 range, since the proto documents no upper bound for them.

One caveat for anyone reading this later: the API *accepts* `decimal_precision = 16` — I applied one. So this guards a documented limit rather than mirroring a backend rejection.

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,554 @@
...
+				Validators: []validator.List{
+					listvalidator.SizeAtLeast(1),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: ObservationFieldSchema(),
+				},
+			},
+			"color_scheme": schema.StringAttribute{
+				Optional: true,
+			},
```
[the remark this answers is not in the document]
dmenilo1: Not taking this one — a `OneOf` here would reject configurations the API accepts. Three things argue against it:

**`colorScheme` is a free-form string, not an enum.** The SDK types it `*string` with `regexp=^[\s\S]*$`, so there is no enum to be exhaustive about.

**Confirmed empirically.** I applied `color_scheme = "not-a-real-scheme"`; the API stored it verbatim and the follow-up plan was clean. A validator would block that, and this repo treats rejecting a config the API accepts as the harder mistake to undo.

**`DashboardValidColorSchemes` is a hardcoded provider-side list**, not derived from an SDK map like the other enum lists here. Nothing keeps it in step with the backend, so a `OneOf` would block any scheme Coralogix adds until someone edits that slice.

On consistency — I checked, and the premise is the other way round. Of 23 `color_scheme` attributes in the package, exactly one is validated: in `linechart.go`, and only one of its two. `common.go` has five unvalidated, the bar visualizations eight, time-series six. Unvalidated is the convention; the validated one is the outlier.

The real problem you have found is that a typo silently applies no scheme and nothing says so. Addressed in 7e7ba90 as documentation instead: the attribute now names the schemes the product ships while still accepting others.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,555 @@
...
+					"show_percentage": schema.BoolAttribute{
+						Optional: true,
+					},
+					"show_value": schema.BoolAttribute{
+						Optional: true,
+					},
+				},
+			},
+			"legend": LegendSchema(),
+			"max_slices_per_chart": schema.Int64Attribute{
```
markshleifer-coralogix, line 209: The proto requires max_slices_per_chart >= 1, max_slices_per_stack >= 1, is it possible to align it?

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,555 @@
...
+					int64validator.Between(0, math.MaxInt32),
+				},
+			},
+			"max_slices_per_stack": schema.Int64Attribute{
+				Optional: true,
+				Validators: []validator.Int64{
+					int64validator.Between(0, math.MaxInt32),
+				},
+			},
+			"min_slice_percentage": schema.Int64Attribute{
```
markshleifer-coralogix, line 223: The proto limits it to a max of 100, is it possible to align it?

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,555 @@
...
+				Validators: []validator.List{
+					listvalidator.SizeAtLeast(1),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: ObservationFieldSchema(),
+				},
+			},
+			"color_scheme": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: fmt.Sprintf("How slices are coloured. The API takes a free-form string and stores whatever it is given, so this is not restricted to a fixed set; the schemes the product ships are: %s. An unrecognised value applies no scheme.", strings.Join(DashboardValidColorSchemes, ", ")),
```
markshleifer-coralogix: What happens if we send a color scheme that is not valid? Where was `DashboardValidColorSchemes` generated from?

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,555 @@
...
+			},
+			"max": schema.Float64Attribute{
+				Optional: true,
+			},
+			"min": schema.Float64Attribute{
+				Optional: true,
+			},
+			"show_inner_arc": schema.BoolAttribute{
+				Optional:            true,
+				MarkdownDescription: "Deprecated: use `arc_display.value_arc` instead.",
```
markshleifer-coralogix, line 108: Please use Terraform’s DeprecationMessage

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,555 @@
...
+			},
+			"show_min_max": schema.BoolAttribute{
+				Optional: true,
+			},
+			"show_outer_arc": schema.BoolAttribute{
+				Optional:            true,
+				MarkdownDescription: "Deprecated: use `arc_display.threshold_arc` instead.",
```
markshleifer-coralogix, line 116: Please use Terraform’s DeprecationMessage

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
dmenilo1: Aligned in `dd5f7eb` — both now require at least 1, with `TestAccCoralogixResourceDashboardDynamicPieRejectsOutOfRangeSliceBounds` covering it.

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
dmenilo1: Aligned in `dd5f7eb` — now `Between(0, 100)`, with a subtest for 101 and -1.

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
dmenilo1: Done in `dd5f7eb` — `DeprecationMessage` now warns at plan time rather than only in the docs.

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
dmenilo1: Done in `dd5f7eb` — same, using `arc_display.threshold_arc`.

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
```diff
@@ -0,0 +1,559 @@
...
+						Optional: true,
+					},
+				},
+			},
+			"legend": LegendSchema(),
+			"max_slices_per_chart": schema.Int64Attribute{
+				Optional: true,
+				Validators: []validator.Int64{
+					int64validator.AtLeast(1),
+				},
```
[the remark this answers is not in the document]
dmenilo1: Good catch, and it was a regression I introduced by taking the minimum-only request literally. Both are back to `Between(1, math.MaxInt32)` in `fcbdeac`, and I added a guard that walks the schema and fails if any int64 attribute lacks an upper bound, since the int32 conversion is unchecked. The same mistake was in #670 across eight attributes; fixed there too.

internal/provider/dashboards/dashboard_widgets/dynamic_gauge_pie.go
dmenilo1: Two answers.

**Invalid value:** nothing rejects it. I applied `color_scheme = "not-a-real-scheme"` against a real environment — the API stored it verbatim and the follow-up plan was clean. The widget just renders with no scheme.

**Provenance:** nowhere authoritative, and I should have checked before citing it. It is a hand-written literal added in #432, unlike its neighbours (`DashboardValidThresholdTypes`, `DashboardValidLegendBys`, …) which are `utils.GetKeys(...)` off an SDK map. There is nothing to derive it from: `colorScheme` is `google.protobuf.StringValue` in the proto and `*string` in the generated client, with no enum model. The proto description does say "one of the predefined values", so a closed set is intended — it just is not expressed anywhere machine-readable, and the backend does not enforce it.

So I left it unvalidated and reworded the description, which previously implied the list was exhaustive: it now says these are the schemes this provider knows about, and that an unrecognised value is stored but applies no scheme.

Happy to switch to a `OneOf` if you would rather catch typos — the trade is that it blocks any scheme added after that list was written. Worth knowing the scope either way: there are 23 `color_scheme` attributes in this package and exactly one is validated today (`linechart.go`, one of its two), so validating here would be new behaviour rather than consistent behaviour, and I would want to do all 23 in one pass. The third option is asking the dashboards team to express the set in the proto, which is the only one that fixes the root cause.

markshleifer-coralogix: I would raise it as a gap to the BE team and won't add validations that are unique to TF and don't exist in the API.
For now I would allow all values since even if a "wrong" value is provided, the widget still renders.

dmenilo1: Agreed, and that is what is in the branch — no validator, all values allowed, and the description now says these are the schemes this provider knows about rather than implying the list is exhaustive.

I will raise the gap with the dashboards team: the proto describes `color_scheme` as "one of the predefined values" but types it as `StringValue`, so there is nothing generated to validate against and `DashboardValidColorSchemes` has stayed a hand-written list since #432. If they express the set as an enum, that list becomes generated like every other one here.

markshleifer-coralogix (approved):

PR #669, opened 2026-08-25
feat(dashboards): add the spatial visualizations to the dynamic widget (CX-50151)

dmenilo1:
## What this does

Adds the last three visualizations to the `dynamic` dashboard widget: `hexagon_bins`, `heatmap` and `geomap`. With #666 this completes all fifteen.

```hcl
visualization = {
  geomap = {
    unit        = "percent"
    aggregation = { count = true }                      # or sum/min/max/avg with a field
    color       = { size = "blue" }                     # or color_range for a gradient
    config = {
      coordinate_config = {                             # or aws_region_config
        latitude_field  = { keypath = ["latitude"], scope = "user_data" }
        longitude_field = { keypath = ["longitude"], scope = "user_data" }
      }
    }
  }
}
```

Branched from `master`, independent of #666 and #670.

## Where to look

| What | Where |
|---|---|
| All three visualizations | `dynamic_geo_heat.go` (new) |
| Models | `common.go` |

**The thing worth checking:** the geomap has three separate unions — where coordinates come from, what each cluster shows, and how clusters are coloured. Each takes exactly one alternative, and one of them is an empty marker rather than a value.

**`aggregation.count` is a `true`/absent switch, not a value**
- *Product:* a cluster either shows how many points it contains, or an aggregate of a field.
- *API:* `GeomapAggregationSimple` is an empty message — `{"count": {}}` on the wire selects counting; the other four arms wrap an observation field.
- *Terraform:* `aggregation = { count = true }` versus `aggregation = { sum = { field = … } }`
- *Why:* a bool is the natural HCL for a marker, and this widget already does exactly this for `mapped_values` and `min_max.auto`.
- *Does the API restrict this too?* Yes, it is a oneof; `ExactlyOneOfChildren` mirrors it, and `count` carries a must-be-true validator because `false` would otherwise satisfy the union while selecting nothing.

**The two colour unions are guarded differently, and the asymmetry is deliberate.** The geomap's `size`/`color_range` pair lives inside a `GeomapColor` message, so it gets `ExactlyOneOfChildren` and setting neither is rejected. The heatmap's `preset`/`color_range` pair sits **directly on the visualization** with no wrapper, so there is no object to attach that validator to — it uses `ConflictsWith` instead, and a heatmap with neither set is legitimate (no colour configuration). Same intent, different shape, because the protos differ.

**`color_axis_min` / `color_axis_max` use the semantic-equality number type** from #648: they are `float32` on the wire, and a `float64` attribute round-tripping through `float32` otherwise produces a permanent diff.

## Scope

**Empty lists are rejected at plan time.** All five lists this surface exposes — the hexagon's `category_fields`, the heatmap's `x_axis_fields` and `y_axis_fields`, and both tooltips' `labels` — had no guard. Without one, `x_axis_fields = []` passes the plan and fails the apply with `Provider produced inconsistent result after apply … was cty.ListValEmpty(…), but now null`, because an empty list and an absent list are different values and the API cannot store the former.

**`decimal_precision` is bounded to 0–15** on all three, matching the proto and the other visualizations. It was at the int32 range in the code this came from.

**The API-coverage manifest is reclassified**, and this is the part worth a reviewer's eye. Three `Visualization` arms move to covered, and the four nested unions — the geomap's field config, aggregation and colour, plus the heatmap colour pair — move out of "reachable only through content_json", since this change gives them a structured path. The arms the tests actually set are marked covered; the geomap's `min`, `max` and `avg` and the heatmap's `preset` are marked as **gaps**, because they share a conversion with a tested arm but nothing sets them. Marking those covered would have been the easier edit and the wrong one.

## Evidence

`TestAccCoralogixResourceDashboardDynamicSpatialWidgets`, against a real environment, on the repo's enforced create/update/import lifecycle: all three visualizations, a fractional colour axis, both tooltips, and a step that removes the optional enums and asserts they fall back. Every step reads the dashboard back from the API, import included — and the backend check asserts the geomap unions stored **exactly one arm each** rather than trusting Terraform state.

`TestDynamicWidgetSpatialFullFidelityRoundTrip` covers write → read across every field of all three, plus a fourth case for the field-based aggregation and AWS-region arms, and runs in PR CI without credentials. I checked it fails when the `count` marker, a tooltip's labels or the `float32` axis is dropped from the read direction.

`…RejectsEmptyLists` covers all five lists. `…RejectsInvalidUnions` covers five cases: two arms and no arm on the geomap aggregation, two arms on the field config, two arms on the colour, and the heatmap preset-plus-range combination.

`TestDynamicWidgetInt32AttributesAreBoundedAbove` is new here and applies to the whole widget: it walks the schema and fails if any int64 attribute lacks an upper bound, because the values reach the API through an unchecked `int32` cast and would otherwise wrap negative between plan and apply. Two string-backed attributes are recorded as exceptions with the reason. It exists because that regression was introduced during review on a sibling PR.

## Verified in the product

Applied against a real account and opened in the UI.

**None of the three is rewritten by the editor** — `hexagonBins`, `heatmap` and `geomap` all survived an edit-and-save. Like `table` and `gauge`/`pie_chart`, they have no sibling variant for the UI's data-shape switching to choose between, so **they carry no release-pairing constraint** and can ship alone.

**The hexagon rendered immediately with one group-by, and needed an editor visit with two.** Same schema, same visualization — only the data changed. That is worth stating because it means "renders on first load" is a property of the data rather than of the visualization, so the earlier per-visualization tally of which ones need a visit was never a real distinction. The underlying behaviour is BUGV2-6905, closed as intended.

**The heatmap needs both axes.** It is a two-dimensional grid, so a configuration with only `x_axis_fields` is accepted by the API and plots nothing. Mentioned because the acceptance tests pass either way — the API is happy to store a configuration that displays nothing.

**Honest limit: the geomap renders its frame but was never seen plotting points.** It needs `latitude`/`longitude` fields, and the account used for verification has no log source carrying coordinates. The configuration round-trips and the widget loads; that the clusters place correctly is untested.

**Expect drift after a UI edit:** the editor fills in optional settings the configuration omits — legend, `allow_abbreviation`, `decimal_precision`, `min_max.auto`, tooltip labels. Ordinary Terraform semantics, since the configuration is authoritative.

## Not included, on purpose

- **`gauge` and `pie_chart`** — #666, which completes the set alongside this.
- **The numeric bounds on already-merged visualizations** — #670.
- **The `Default: UNSPECIFIED` question.** Around 119 attributes across `internal/provider/dashboards/` reset to `unspecified` when omitted, so anything the UI writes flips back on the next apply. Pre-existing and repo-wide; its own PR, once it is clear which of them the UI actually writes.


dmenilo1:
@codex review

internal/provider/dashboard_openapi_oneof_coverage_test.go
```diff
@@ -242,6 +243,9 @@ func dashboardOpenAPIOneOfCoverageManifest() map[string]dashboardOneOfModelCover
 	visualization.Branches["timeSeriesLines"] = covered(dynamicWidget+".visualization.time_series_lines", dashboardOpenAPIDynamicTimeSeriesTestName)
 	visualization.Branches["timeSeriesLinesMulti"] = covered(dynamicWidget+".visualization.time_series_lines_multi", dashboardOpenAPIDynamicTimeSeriesTestName)
 	visualization.Branches["timeSeriesBars"] = covered(dynamicWidget+".visualization.time_series_bars", dashboardOpenAPIDynamicTimeSeriesTestName)
+	visualization.Branches["hexagonBins"] = covered(dynamicWidget+".visualization.hexagon_bins", dashboardOpenAPIDynamicGeoHeatTestName)
+	visualization.Branches["heatmap"] = covered(dynamicWidget+".visualization.heatmap", dashboardOpenAPIDynamicGeoHeatTestName)
+	visualization.Branches["geomap"] = covered(dynamicWidget+".visualization.geomap", dashboardOpenAPIDynamicGeoHeatTestName)
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `c9f9d65` — all four are now branch entries rather than `apiOnlyModel`. The arms the tests set are marked covered; the geomap `min`/`max`/`avg` and the heatmap `preset` are marked as gaps, since they share a conversion with a tested arm but are not exercised themselves.

dmenilo1:
@codex review

internal/provider/dashboard_openapi_oneof_coverage_test.go
```diff
@@ -389,26 +395,39 @@ func dashboardOpenAPIOneOfCoverageManifest() map[string]dashboardOneOfModelCover
...
-		"Heatmap": apiOnlyModel(
-			"ast/widgets/dynamic.proto#Dynamic.Visualization.PropertyLinks.Heatmap.color_config",
-			"heatmap color configuration is reachable only below WidgetDefinition.dynamic",
-			"preset", "colorRange",
-		),
+		"GeomapAggregation": {
+			ProtoSource: "ast/widgets/dynamic.proto#Dynamic.Visualization.PropertyLinks.Geomap.GeomapAggregation.value",
+			Branches: map[string]dashboardOneOfBranchCoverage{
+				"count": covered(geomap+".aggregation.count", dashboardOpenAPIDynamicGeoHeatTestName),
+				"sum":   covered(geomap+".aggregation.sum", dashboardOpenAPIDynamicGeoHeatTestName),
```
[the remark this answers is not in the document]
dmenilo1: Addressed in `3720fc6` — rather than downgrade them, the acceptance lifecycle now applies a second geomap widget carrying `sum`, `color_range` and `aws_region_config`, with backend assertions that each union stored exactly one arm. So all three are covered through apply, read and import, and the classification is accurate. `min`, `avg` and the heatmap `preset` remain gaps: nothing applies them.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1068 @@
...
+			"category_fields": schema.ListNestedAttribute{
+				Optional: true,
+				Validators: []validator.List{
+					listvalidator.SizeAtLeast(1),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: ObservationFieldSchema(),
+				},
+			},
+			"custom_unit": schema.StringAttribute{
```
markshleifer-coralogix, line 105: The API defines custom_unit with a length of 1–128 characters. This schema has no length validator. The same gap exists in the heatmap and geomap custom_unit attributes below.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1068 @@
...
+			"labels": schema.ListNestedAttribute{
+				Optional: true,
+				Validators: []validator.List{
+					listvalidator.SizeAtLeast(1),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: ObservationFieldSchema(),
+				},
+			},
+			"message_template": schema.StringAttribute{
```
markshleifer-coralogix, line 269: The API defines tooltip message_template with a length of 1–4096 characters. This schema has no length validator. The same gap exists in the geomap tooltip.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1068 @@
...
+// Schemas --------------------------------------------------------------------
+
+func dynamicHexagonBinsSchema() schema.Attribute {
+	return schema.SingleNestedAttribute{
+		Optional: true,
+		Attributes: map[string]schema.Attribute{
+			"allow_abbreviation": schema.BoolAttribute{
+				Optional: true,
+			},
+			"category_fields": schema.ListNestedAttribute{
```
markshleifer-coralogix, line 96: The API caps these lists at 1000 items. The current schemas only reject empty lists and do not enforce the upper bound. Same for both axis lists, both tooltip label lists, and thresholds.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
dmenilo1: Confirmed the limit is real but not enforced — I applied a 129-character `custom_unit` and the API stored all 129. Doing it as a follow-up across all fifteen visualizations rather than only these three, since `custom_unit` appears on nearly every one and there are currently no length validators anywhere in `dashboards/`.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
dmenilo1: Also confirmed unenforced — 1001 tooltip labels applied and stored. Same follow-up: the proto documents 46 of these caps and the package enforces none today, so a single pass beats three edits here.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
dmenilo1: Same — a 4097-character `message_template` applied and stored verbatim. Rolling it into the same follow-up as `custom_unit`, covering every visualization at once.

markshleifer-coralogix (dismissed):

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
dmenilo1: Following up on this — the survey turned up two things that change how I would scope it, so flagging before I write the follow-up PR.

**The limit is not uniform.** `custom_unit` is `max_length: 128` in `dynamic.proto` but **255** in every legacy widget proto (bar_chart, gauge, hexagon, horizontal_bar_chart, line_chart, pie_chart). So the obvious design — one shared `CustomUnitSchema()` helper alongside the existing `UnitSchema()` — would be wrong: it would impose 128 on legacy widgets whose API allows 255.

**Release exposure differs sharply.** 15 of the 17 declarations are in the dynamic widget and all but one are unreleased. The 2 in `hexagon.go` have shipped for years, and since the API does not enforce the limit, adding a 255 validator there could break a live configuration on upgrade with no correctness gain.

So unless you would rather otherwise, I plan to scope the follow-up to the **dynamic widget only**: `custom_unit` 1–128 across its 15 declarations, `message_template` 1–4096 on both, and `SizeAtMost(1000)` on its lists. Legacy widgets left alone and called out in the description, so covering them stays a deliberate decision with a changelog warning rather than a side effect.

One honest caveat inside that scope: `stat` shipped its `custom_unit` in 3.10.0, six days ago, so a 129-character value is technically possible in the wild. Given the age and that it is a unit label, the risk looks negligible, but I will note it in the changelog rather than leave it implicit.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1071 @@
...
+}
+
+func expandDynamicMinMax(m *DynamicMinMaxModel) *dashboardservice.MinMax {
+	if m == nil {
+		return nil
+	}
+	minMax := &dashboardservice.MinMax{}
+	if m.Auto.ValueBool() {
+		minMax.Auto = map[string]interface{}{}
+	}
```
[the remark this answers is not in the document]
dmenilo1: Right, and my own fix created the failure mode — normalising the empty wrapper away on read is what turned this into an inconsistent-result error. Fixed in `c80a7c4` using the same diagnostic as `expandDynamicRangeMapping`, with a test that also asserts both valid arms still convert.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1084 @@
...
+
+	aggregation := &dashboardservice.GeomapAggregation{
+		Avg: avg,
+		Max: maxAgg,
+		Min: minAgg,
+		Sum: sum,
+	}
+	if m.Count.ValueBool() {
+		aggregation.Count = map[string]interface{}{}
+	}
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `b111ecf`. Rather than patch just this one, I enumerated every union in the file: five, and only `min_max` re-checked. The geomap aggregation, colour and field config now do too — the latter two have the same structure and were not flagged, so they are fixed here rather than one review round at a time.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1084 @@
...
+
+	return &dashboardservice.Heatmap{
+		AllowAbbreviation:   m.AllowAbbreviation.ValueBoolPointer(),
+		ColorAxisMax:        expandFloat32Pointer(m.ColorAxisMax),
+		ColorAxisMin:        expandFloat32Pointer(m.ColorAxisMin),
+		ColorRange:          OptionalEnumPointer(m.ColorRange, dashboardSchemaToProtoColorGradientType),
+		CustomUnit:          m.CustomUnit.ValueStringPointer(),
+		DecimalPrecision:    expandInt32Pointer(m.DecimalPrecision),
+		HistogramBucketUnit: OptionalEnumPointer(m.HistogramBucketUnit, dashboardSchemaToProtoHeatmapHistogramBucketUnit),
+		Preset:              OptionalEnumPointer(m.Preset, dashboardSchemaToProtoHeatmapColorPreset),
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `b111ecf`. The heatmap is the one that differs: neither colour arm set is legitimate, since a heatmap need not be coloured, so only the both-set case is refused — the test asserts the empty case still converts.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -903,6 +909,31 @@ func expandDynamicVisualizationFamilyGaugePie(ctx context.Context, visualization
 	return nil, nil
 }
 
+func expandDynamicVisualizationFamilySpatial(ctx context.Context, visualization *DynamicVisualizationModel) (*dashboardservice.Visualization, diag.Diagnostics) {
+	switch {
+	case visualization.HexagonBins != nil:
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `b4b893e`. The arms are counted by reflection rather than a hand-written list, so a visualization added to the model cannot drift out of the check — the test asserts that too.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1127 @@
...
+// object validator cannot catch this when `auto` is only known after apply, so
+// the conversion has to refuse it.
+func expandDynamicMinMax(m *DynamicMinMaxModel) (*dashboardservice.MinMax, diag.Diagnostics) {
+	if m == nil {
+		return nil, nil
+	}
+
+	switch {
+	case m.Custom != nil:
+		return &dashboardservice.MinMax{Custom: &dashboardservice.MinMaxCustom{
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `b4b893e` — my earlier fix handled zero arms and left two, and the test only covered zero. Also fixed the same gap in `expandDynamicRangeMapping`, which is the code this was modelled on and had it too.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -779,10 +786,20 @@ func expandDynamicVisualization(ctx context.Context, visualization *DynamicVisua
 		return nil, nil
 	}
 
+	// ExactlyOneOfChildren defers while any arm is unknown, so a value only
+	// known after apply can arrive with two visualizations set. The dispatch
+	// below takes the first and would silently discard the rest.
+	if selected := dynamicSelectedVisualizations(visualization); len(selected) > 1 {
```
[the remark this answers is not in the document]
dmenilo1: Fixed — it now requires exactly one. Same asymmetry as my `min_max` fix, inverted: there I handled zero and missed two, here I handled two and missed zero.

Rather than wait for the next one, I completed the matrix across all six unions and added the two cases that were guarded but untested: the geomap field config with no arm, and the stat card range mapping with both. The heatmap colour stays the deliberate exception, since a heatmap need not be coloured — 14 subtests now cover it.

dmenilo1:
@codex review

dmenilo1:
@codex review

internal/provider/resource_coralogix_dashboard_dynamic_geo_heat_test.go
```diff
@@ -0,0 +1,423 @@
...
+	// trusting Terraform state.
+	geomap := rows[1].GetWidgets()[0].Definition.Dynamic.Visualization.Geomap
+	if geomap == nil || geomap.Aggregation == nil {
+		return fmt.Errorf("dashboard fixture %q (dashboard %q): geomap has no aggregation", fixture, dashboard.GetId())
+	}
+	if geomap.Aggregation.Count == nil {
+		return fmt.Errorf("dashboard fixture %q (dashboard %q): geomap aggregation count marker was not stored", fixture, dashboard.GetId())
+	}
+	if geomap.Aggregation.Sum != nil || geomap.Aggregation.Avg != nil {
+		return fmt.Errorf("dashboard fixture %q (dashboard %q): geomap aggregation stored more than one arm", fixture, dashboard.GetId())
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `c56b75c`, and it was a bit worse than you describe: the first fixture also never checked the colour union at all, and the second never checked its coordinate config. Each union is now asserted over all of its arms — exactly one set, and it is the expected one — rather than spot-checking a couple.

One thing the mutation test turned up, worth recording: a converter setting two arms **cannot** reach this assertion, because the SDK's `MarshalJSON` refuses two arms of a oneof before the request is sent. So the assertion guards the read path — a backend returning two arms or none — which is not reachable from a converter mutation. It is unit-tested directly instead, covering two arms, no arm, the wrong arm, and the union missing entirely.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1118 @@
...
+
+	aggregation := &dashboardservice.GeomapAggregation{
+		Avg: avg,
+		Max: maxAgg,
+		Min: minAgg,
+		Sum: sum,
+	}
+	if m.Count.ValueBool() {
+		aggregation.Count = map[string]interface{}{}
+	}
```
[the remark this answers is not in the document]
dmenilo1: Fixed in `fc49910`, at the source rather than here. Both of these hand-rolled the marker conversion, so they now use the shared helper — which itself now returns a diagnostic for an explicit false. That means the same gap in the stat card's `mapped_values`, its template variables and the bar sort strategy is closed too; three more sites you had not flagged.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1118 @@
...
+
+	switch {
+	case m.Custom != nil && m.Auto.ValueBool():
+		return nil, dynamicUnionDiagnostic("min_max", "`auto` set to true or `custom`, not both")
+	case m.Custom != nil:
+		return &dashboardservice.MinMax{Custom: &dashboardservice.MinMaxCustom{
+			Max: m.Custom.Max.ValueFloat64Pointer(),
+			Min: m.Custom.Min.ValueFloat64Pointer(),
+		}}, nil
```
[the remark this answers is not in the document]
dmenilo1: Same fix in `fc49910` — `min_max.auto` goes through the shared marker conversion now, so a false auto is refused whether or not `custom` is selected.

One note on verifying it: my first mutation made false return a non-nil marker, which the exactly-one check then caught for the wrong reason, so only 1 of 4 subtests failed. Restoring the true silent-drop behaviour fails all 4, which is what actually proves the guard.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1125 @@
...
+	if diags.HasError() {
+		return nil, diags
+	}
+
+	count := types.BoolNull()
+	if m.Count != nil {
+		count = types.BoolValue(true)
+	}
+
+	return &DynamicGeomapAggregationModel{
```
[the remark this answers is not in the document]
dmenilo1: Fixed. I audited every union flatten in the widget rather than the three you named, and two more had it: the bar chart's `sort_order.strategy` and the stat card's `color_label_mapping`. All six normalise now.

Each guard has exactly one test that fails without it — verified individually, no overlap — plus a case asserting a populated wrapper still survives, so a guard cannot pass by dropping the field instead.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1139 @@
...
+
+func expandDynamicHexagonBins(ctx context.Context, m *DynamicHexagonBinsModel) (*dashboardservice.HexagonBins, diag.Diagnostics) {
+	if m == nil {
+		return nil, nil
+	}
+
+	categoryFields, diags := ExpandObservationFields(ctx, m.CategoryFields)
+	if diags.HasError() {
+		return nil, diags
+	}
```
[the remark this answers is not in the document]
dmenilo1: Confirmed, and it does not depend on how the API treats an empty list: flatten reads any empty list back as null, so config `[]` versus state null is the inconsistency on its own.

Fixed one level up rather than at the five sites. All 120 min-validated lists in the widget have the same hole, so conversion now walks the model once and rejects a known-empty list wherever it appears, naming the attribute. A list added later is covered without anyone remembering to add a check.

The two lists where empty means everything keep round-tripping empty, and unknown and null values still pass through — only a known-empty list is a mistake. Four mutations of the walk each fail a test, including one asserting it descends into list elements rather than only the lists hanging off the model.

internal/provider/dashboards/dashboard_widgets/dynamic_stat_card.go
```diff
@@ -690,6 +714,11 @@ func expandDynamicRangeMapping(ctx context.Context, rangeMapping *DynamicRangeMa
 	var minMax *dashboardservice.MinMax
 	if rangeMapping.MinMax != nil {
 		switch {
+		case rangeMapping.MinMax.Custom != nil && rangeMapping.MinMax.Auto.ValueBool():
```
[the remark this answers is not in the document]
dmenilo1: Confirmed. The cause is that this was a second copy of the min/max conversion, written before the marker rules existed, so the rules never reached it — both directions were duplicated here.

Fixed by deleting the copy rather than patching it: expand and flatten now call the shared helpers, 40 lines down to 4, and there is one place these rules live.

Swapping the old logic back in fails `false auto beside custom` and nothing else, which is exactly the hole you identified — the arm-less and both-arms cases were already covered by its `default` branch. Also asserted a backend `min_max` with neither arm reads back unset.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1139 @@
...
+	tooltip, diags := expandDynamicHeatmapTooltip(ctx, m.Tooltip)
+	if diags.HasError() {
+		return nil, diags
+	}
+
+	// Unlike the other unions here, neither arm set is legitimate: a heatmap
+	// with no colour configuration is valid. Only both at once is impossible.
+	colorRange := OptionalEnumPointer(m.ColorRange, dashboardSchemaToProtoColorGradientType)
+	preset := OptionalEnumPointer(m.Preset, dashboardSchemaToProtoHeatmapColorPreset)
```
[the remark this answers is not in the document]
dmenilo1: Confirmed — `OptionalEnumPointer` returns nil on a mapping miss, so the attribute is dropped and the read flattens it to null.

Rather than diagnose these two, conversion now replays the schema's own validators against the resolved values. The same hole covers 150 enums, 44 numeric bounds and 120 list sizes; replaying the declared rules avoids restating any of them in a second place, and it also replaced the hand-written empty-list check from earlier on this branch.

One thing worth flagging from doing it this way: only a validator that judges its own value can be replayed. One that inspects a sibling needs the configuration, and without one it reports "Invalid Path Expression for Schema" on a perfectly valid dashboard — I nearly shipped exactly that. `ConflictsWith`, `ExactlyOneOfChildren` and the markers therefore stay checked by hand where the arm is chosen, and a test asserts every validator in the schema is classified as one or the other so a new one cannot be skipped silently. 359 instances classified today.

Four mutations each fail a test, including replaying `ConflictsWith`, which breaks a valid heatmap that sets only `color_range`.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -1424,3 +1630,110 @@ func flattenOptionalEnum[T ~string](value *T, mapping map[T]string) types.String
...
+	filtered := make([]V, 0, len(validators))
+	for _, candidate := range validators {
+		if dynamicValueOnlyValidators[fmt.Sprintf("%T", candidate)] {
+			filtered = append(filtered, candidate)
+		}
+	}
+	return filtered
+}
+
+func dynamicRunValidators(ctx context.Context, value attr.Value, attribute schema.Attribute, at path.Path) diag.Diagnostics {
```
markshleifer-coralogix: This replay handles scalar and list validators only. It does not replay boolean or object validators. Can we add conversion checks or replay these validators with the full object value?

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1139 @@
...
+			"message_template": schema.StringAttribute{
+				Optional: true,
+			},
+		},
+	}
+}
+
+func dynamicGeomapSchema() schema.Attribute {
+	return schema.SingleNestedAttribute{
+		Optional: true,
```
markshleifer-coralogix, line 278: The proto describes config as required geographic information and says aggregation should default to count. The current schema allows geomap = {} and sends neither value. Should these be required, or should the provider set a default count?

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1139 @@
...
+			"histogram_bucket_unit": schema.StringAttribute{
+				Optional: true,
+				Computed: true,
+				Default:  stringdefault.StaticString(utils.UNSPECIFIED),
+				Validators: []validator.String{
+					stringvalidator.OneOf(dashboardValidHeatmapHistogramBucketUnit...),
+				},
+				MarkdownDescription: fmt.Sprintf("The histogram bucket unit. Valid values are: %s.", strings.Join(dashboardValidHeatmapHistogramBucketUnit, ", ")),
+			},
+			"preset": schema.StringAttribute{
```
markshleifer-coralogix, line 200: HeatmapColorPreset is deprecated in the proto. Should we add deprecation message for it in preset and recommend color_range?

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -561,11 +568,111 @@ func dynamicThresholdAttr() map[string]attr.Type {
...
 }
 
+// A validator returns early when its value is unknown, and validators do not
+// run again at apply time, so anything written from a variable, a data source
+// or another resource reaches conversion with nothing having checked it. An
+// invalid value is then silently dropped or truncated and the apply fails with
+// an inconsistent-result error naming nothing in particular. Re-running the
+// schema's own validators against the resolved values covers every attribute at
+// once, including ones added later, and keeps a single definition of each rule.
+func dynamicRevalidateResolved(ctx context.Context, dynamic *DynamicModel) (int, diag.Diagnostics) {
```
markshleifer-coralogix: This introduces a second validation engine for the dynamic widget. It walks framework schemas with reflection, invokes validators outside the framework, and classifies validators by formatted concrete type names. This is a new provider pattern and affects all existing dynamic visualizations.
Terraform documents that configuration validation is eventually called with known values during apply. IS it really needed? It's a new paradigm that was introduced to the repo.

internal/provider/dashboards/dashboard_widgets/dynamic.go
```diff
@@ -796,6 +920,38 @@ func expandDynamicVisualization(ctx context.Context, visualization *DynamicVisua
 	return nil, nil
 }
 
+// Counted by reflection rather than a hand-written list, so a visualization
+// added to the model cannot drift out of the check above.
+func dynamicSelectedVisualizations(visualization *DynamicVisualizationModel) []string {
```
markshleifer-coralogix: Please avoid reflection for this static union. Use an explicit typed list such as {name, selected} and add a schema-wiring test that compares its names with the visualization attributes. This keeps compile-time references to every model field and follows the existing ExactlyOneOfChildren drift-test pattern. Reflection also assumes that every future pointer field in this model is a visualization arm.

internal/provider/dashboards/dashboard_widgets/dynamic.go
dmenilo1: Verified both halves before answering.

The gap is real: I set a `OneOf` attribute from another resource's computed `id` — unknown at plan, an invalid string at apply — and no validation error fired. Validators do not re-run with resolved values.

But you are right that a documented hook with that guarantee exists, and I picked the wrong one. `ModifyPlan` is documented as called "once during the plan phase ... and once during the apply phase with any unknown values from configuration filled in with their final values", while `ValidateConfig` only promises config that "may contain unknown values".

So I am moving the check into `ModifyPlan`. That drops the reflection walk and the type-name classification entirely, since `ModifyPlan` receives the full `tfsdk.Config` — which also fixes your bool/object comment. Standard interface rather than a new paradigm. Thanks, this is a better design than what I wrote.

internal/provider/dashboards/dashboard_widgets/dynamic.go
dmenilo1: Same root cause as the ModifyPlan comment — bool and object validators were excluded because they are the marker (`mustBeTrue`, re-checked by hand in expand) and `ExactlyOneOfChildren`, which needs `req.Config` and emits a bogus "Invalid Path Expression for Schema" on valid config without it. Moving the replay into `ModifyPlan` gives us the real `Config`, so all kinds replay uniformly and the allowlist goes away.

internal/provider/dashboards/dashboard_widgets/dynamic.go
dmenilo1: Agreed, and it is worse than a style issue — treating every non-nil pointer as an arm means the next non-visualization pointer added to that model silently counts as a second arm and breaks every config. Switching to an explicit typed list plus a drift test against the schema attributes, matching the `ExactlyOneOfChildren` pattern.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
dmenilo1: Checked the spec rather than the proto comment, and it says optional. `Geomap` declares no `required` list, and the spec does use `required` where it means it — 17 of 396 schemas do, including `Dashboard` and `Dynamic.QueryDefinition`. No schema in the spec declares a `default` at all (0 of 396), so "should always be set to count" is prose, not an enforced default. The live API agrees: an empty geomap is accepted and stored, and no count is filled in.

Making `config` required would reject stored dashboards the API returns, and a provider-side count default would send a value the user never wrote and diff forever on geomaps stored without one. Putting the intent into the attribute descriptions instead.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
dmenilo1: Cannot find it in the SDK we pin (v1.9.4-0.20260824133534). The generator does emit deprecation — `XAxis.Time` carries `// Deprecated` — but neither the `preset` field, the `HeatmapColorPreset` type, nor `openapi.yaml` has any marker. Could you point me at the proto line? Happy to add it, just want to confirm first since a `DeprecationMessage` shows up in every user's plan output.

dmenilo1: Dug into this properly. The spec marks exactly 20 fields `deprecated: true`, and `Heatmap.preset` is not one of them — nor is `colorRange`, so nothing in the spec suggests one supersedes the other either. Generated Go agrees.

One caveat that makes your claim plausible rather than wrong: all 20 markers are on **properties**, and zero of the 396 schemas carry a type-level `deprecated`. So if the proto deprecates the enum type `HeatmapColorPreset` — which is what you named — the OpenAPI generator would not carry it and I have no way to see it from the SDK. That is a blind spot on our side, not evidence against you.

For what it is worth, the deprecations we do know about are already marked: `interpretation`, `time_series_lines`, `show_inner_arc`, `show_outer_arc`, `value_field`, `variables`, `appearance.width`. So the practice is in place; this field just is not in the set the SDK exposes.

Point me at the proto line and I will add `DeprecationMessage` recommending `color_range` — it is one line. I only wanted to confirm first, since a deprecation notice shows up in every user's plan output.

Separately I would like to add a guard test that cross-checks the spec's `deprecated: true` fields against our `DeprecationMessage` coverage, so a future SDK bump that deprecates something we expose fails loudly. Would land as its own change rather than here.

internal/provider/dashboards/dashboard_widgets/dynamic_geo_heat.go
```diff
@@ -0,0 +1,1139 @@
...
+}
+
+// Expand ---------------------------------------------------------------------
+
+func expandDynamicHexagonBins(ctx context.Context, m *DynamicHexagonBinsModel) (*dashboardservice.HexagonBins, diag.Diagnostics) {
+	if m == nil {
+		return nil, nil
+	}
+
+	categoryFields, diags := ExpandObservationFields(ctx, m.CategoryFields)
```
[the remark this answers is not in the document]
dmenilo1: Correct that the revert removes the replay — that was deliberate, and the sequencing is the point rather than an oversight.

The human reviewer objected that the replay was a bespoke validation engine sitting inside a PR that adds three visualizations, and they were right on the mechanism too: the framework documents `ModifyPlan` as called a second time during apply with unknown values resolved, which is exactly the guarantee I hand-rolled. So the replacement is not a relocation of that code, it is a smaller piece of it, and it applies to the whole dashboard resource rather than the dynamic widget alone.

On scope: master has 43 min-size-validated list declarations and zero conversion-time replay, so this hole is provider-wide and pre-dates the branch. The spatial lists here are new instances of it, not a new defect class — the tip is at parity with master, not behind it. Fixing it per-list is what produced 43 unguarded lists in the first place.

Tracking the `ModifyPlan` change as its own PR, which is where this and the primary-value marker case both get fixed together.

internal/provider/dashboards/dashboard_widgets/dynamic_stat_card.go
```diff
@@ -362,8 +362,13 @@ func expandDynamicStatVisualElement(ctx context.Context, element *DynamicStatVis
 		return nil, diags
 	}
 
+	mappedValues, diags := expandDynamicMappedValuesMarker("mapped_values", element.MappedValues)
```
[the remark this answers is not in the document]
dmenilo1: Checked this one against master rather than assuming, and it is not a regression: master's `expandDynamicStatVisualElement` calls the marker conversion with no primary-value restriction either. The revert restored that behaviour exactly, so a variable-derived `primary_value.mapped_values` reaches the API on master today and on this tip identically.

Agreed the restriction needs to survive into apply. That belongs with the same `ModifyPlan` change as the list comment, since `unsupportedMappedValuesValidator` is one of the validators that needs the full configuration to run — replaying it outside the framework is what forced the type-name allowlist the human reviewer objected to. Doing it in `ModifyPlan` gives it the real config and drops the allowlist.

dmenilo1:
@markshleifer-coralogix 
  Two commits since your review: the apply-time revalidation is reverted, and the reflection union counter is replaced with an explicit list plus a drift test.

  **On the revalidation** — the gap is real (verified: a `OneOf` attribute set from another resource's computed `id` is never re-checked at apply), but you were right
  about the mechanism and I'd picked the wrong hook. `ModifyPlan` is documented as running a second time during apply with unknown values resolved; `ValidateConfig`
  only promises config that "may contain unknown values". Doing it there drops the reflection walk and the validator-type allowlist entirely, since the hook gets the
  full `tfsdk.Config`. Reverted rather than reworked here, because it changes validation for every dynamic visualization and shouldn't ride inside a PR adding three of
  them. Own PR to follow.

  **Geomap `config` / `aggregation`** — checked the spec, not the proto comment. `Geomap` declares no `required` list, and the spec does use `required` where it means
  it (17 of 396 schemas). No schema declares a `default` anywhere, so "should always be set to count" is prose. The API confirms it — an empty geomap is stored as-is
  with no count added. Making `config` required would reject dashboards the API returns; a provider-side default would diff forever on geomaps stored without one.
  Intent goes into the attribute descriptions instead.

  **The two Codex comments about the revert** — both checked against master. Lists: master has 43 min-size-validated list declarations and zero conversion-time replay,
  so the hole is provider-wide and pre-dates this branch; this tip is at parity, not behind. Primary-value marker: not a regression — master's
  `expandDynamicStatVisualElement` has no restriction either. Both get fixed by the `ModifyPlan` change.

markshleifer-coralogix (approved):

PR #685, opened 2026-08-27
fix(dashboards): read a bar chart x-axis that groups by time (CX-50156)

dmenilo1:
Closes CX-50156.

### Description

A bar chart whose x-axis groups by time could not be read at all:

```
Error: Error Flatten BarChart XAxis
unknown bar chart x axis type
```

`plan`, `refresh`, `apply` and `import` all failed against that dashboard until the provider was fixed. Reproduced end to end on a dashboard built by hand in the Coralogix UI with no unusual settings — three clicks produce it.

**It reaches dashboards that Terraform created.** The UI writes the x-axis interval as `timeBuckets`, and rewrites a stored `time` block to it whenever the dashboard is saved, including a save that changes only a colour. So a managed bar chart became unmanageable the moment anyone opened it in the UI. An x-axis stored with no kind selected — which the API also accepts — hit the same branch.

The UI's migration itself is faithful: `bucketsPresented: 7` becomes `maximumDataPoints: 7`, and an explicit interval carries over as a manual interval. Nothing is lost on their side. The defect was entirely the read.

### Deviations, and why

**From the API contract.** `IntervalResolution` looks like a one-of but is not. Probed against a live environment:

| Sent | Result |
|---|---|
| `auto` only, `manual` only | accepted |
| neither arm, `{}` | **accepted and stored** |
| `useAdvancedLimit` alone | accepted and stored |
| both arms | rejected when the request is marshalled |

So the rule is **at most one**, never "exactly one". `ExactlyOneOfChildren` would have rejected a resolution the API stores and returns, leaving those dashboards unmanageable — the same mistake I nearly made on CX-50155 by assuming the legacy and new fields were mutually exclusive. That needed a new house validator, `AtMostOneOfChildren`, sitting beside the existing `ExactlyOneOfChildren` and `AtLeastOneOfChildren`.

Anything storable must also be readable, so a resolution or an x-axis with nothing selected reads back as **absent** rather than as a block whose children are all null — state no configuration can express, which would diff on every plan.

**From the repo's other duration fields, deliberately.** The provider already has three duration formats and none round-trips safely: `seconds:900` accepts `minutes:15` but returns `seconds:900`; `1h0m5s` accepts `5m` but returns `5m0s`. Both leave a permanent "pending change" for a reasonable-looking value, and the bar chart's existing validator even suggests `1m`, one of the forms that drifts.

These attributes are therefore sent and read back **without conversion**, in the form the API returns, so there is nothing to get wrong. Verified value by value:

```
900s  15s  0s  86400s  1.500s  0.250s  1.000000001s     returned verbatim
1.5s -> 1.500s     1.0s -> 1s     900.000s -> 900s      rewritten by the API
```

The validator derives the canonical form by re-encoding rather than matching a pattern, so it can name the string to write: `1.5s` is reported as *stored by the API as "1.500s"*. Copy-paste from a UI JSON export works as-is.

**`use_advanced_limit` is a plain `Optional` bool**, not `Optional + Computed`. The API does not fill it in when omitted — verified — so there is nothing to reconcile and no first-plan diff, unlike the enums in #680 which needed `UseNonNullStateForUnknown` precisely because the API does choose a value there. It is a UI editability flag and does not gate the limit fields.

### Scope: feature work versus drive-by

Backward compatible. `xaxis.time` and `xaxis.value` keep working untouched, and no existing configuration changes.

**One new warning, which is the only user-visible change for existing configurations.** `xaxis.time` now carries a deprecation:

```
Warning: Attribute Deprecated

  with coralogix_dashboard.example,
  on main.tf line 12, in resource "coralogix_dashboard" "example":

The Coralogix UI writes `time_buckets` and rewrites this block to it when the
dashboard is saved, so a configuration using `time` can stop matching the
dashboard it manages. Use `time_buckets` instead.
```

It appears on `plan` and `apply`, and repeats until the configuration migrates. The framework emits it only for a set, known value:

```go
if a.GetDeprecationMessage() != "" && !configHasNullValue && !configHasUnknownValue {
```

so it is silent for a configuration using `value` or `time_buckets`, and silent for one that omits the block. Nothing retroactive — only this version onward.

Who actually sees it is a narrow group: a bar chart that came from the UI already stores `timeBuckets`, and an imported one gets `time_buckets` in its configuration. It is only hand-written configurations still using the older block — which are exactly the ones at risk of the silent drift above. `time_buckets.manual.interval` is a direct substitute for `time.interval`.

The schema, model, converters and duration validator are shared rather than bar-chart-local, because CX-50155 needs the same interval resolution on line chart query definitions and annotation metrics sources.

### Evidence

**The read regression, proven rather than asserted.** `TestAccCoralogixResourceDashboardBarChartReadsEveryStoredXAxis` creates each stored shape through the API and reads it back through the provider. Run against a clean `master` worktree, **five of its six cases fail** and `the deprecated time bucket` passes — so it tests the fix and does not regress what already worked.

Also run against the real UI-built dashboard `9IdMP6oskKprT5Fd0tMAW`, stored as `{"timeBuckets":{"auto":{"minimumInterval":"15s"}}}`: fails on `master` with the error above, passes here. That shape is what the `time buckets auto` fixture reconstructs, so the matrix matches what the product writes rather than what I assumed.

**Lifecycle.** `TestAccCoralogixResourceDashboardBarChartTimeBuckets` applies `auto`, import-verifies, switches to `manual` with `use_advanced_limit`, then switches to the deprecated `time`, asserting the previous arm is gone at each step.

**Rejections.** Both interval kinds together, both resolution modes together, no kind at all, and two bad duration forms.

**Unit.** The duration validator over every value probed above including null and unknown; `AtMostOneOfChildren` over none, one, two and the unknown cases; the empty-wrapper normalisation, with a case asserting a resolution carrying only `use_advanced_limit` still reads back so the guard cannot pass by dropping the field.

**Migration.** The registry-backed upgrade from provider v3.6.0 passes, including its `gauge-and-bars` case. Extending the shared `BarChartXAxisModel` was a real risk — the V2 state upgrader decodes prior state into it — so I tested it rather than assuming, and the frozen v1–v3 schemas are untouched because each declares `xaxis` inline.

### Acceptance tests

```
$ TF_ACC=1 go test ./internal/provider -run '^TestAccCoralogix(Resource|DataSource)Dashboards?'
ok      github.com/coralogix/terraform-provider-coralogix   556.209s
```

Build, vet, the full unit sweep and gocyclo are clean, and `make generate` produces no further drift.

### Not included, on purpose

`xaxis.time` is not removed. It still reads dashboards that store it, and removing it would break configurations for no gain.

The line chart and annotation interval resolutions are CX-50155. This lands first because it is a hard read failure that a stock UI produces, where the line chart silently drops the value instead.

The classic widgets' other duration fields keep their existing formats and their existing round-trip traps. Fixing those is a separate change with its own compatibility question.

### Release Note

```release-note
resource/coralogix_dashboard: A bar chart whose x-axis groups by time is readable again, and the x-axis takes `time_buckets` with `auto` or `manual` interval resolution. The Coralogix UI writes this field and rewrites a stored `time` block to it on any save, after which the provider could not read the dashboard at all.
```


internal/provider/dashboards/dashboard_widgets/interval_resolution.go
```diff
@@ -0,0 +1,224 @@
...
+			MinimumInterval:   types.StringPointerValue(resolution.Auto.MinimumInterval),
+		}
+	case resolution.Manual != nil:
+		model.Manual = &ManualIntervalResolutionModel{
+			Interval:          types.StringValue(resolution.Manual.Interval),
+			MaximumDataPoints: int32PointerToInt64Type(resolution.Manual.MaximumDataPoints),
+			MinimumInterval:   types.StringPointerValue(resolution.Manual.MinimumInterval),
+		}
+	case model.UseAdvancedLimit.IsNull():
+		return nil
```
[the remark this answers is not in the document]
dmenilo1: Confirmed by reproducing it — `time_buckets = {}` failed with `Provider produced inconsistent result after apply`. Fixed: a resolution with no mode is kept now, and only a kindless x-axis reads back as absent.

The mistake was carrying a normalisation across without checking the validator differed. On the geomap an arm-less block *had* to become absent because `ExactlyOneOfChildren` rejects an all-null block, so no config could express it. Here `AtMostOneOfChildren` allows none, so `time_buckets = {}` is writable and must survive the read.

Root cause on my side: the read matrix creates each shape through the API and reads it back, which proves readability but never round-trip. Added a test that applies all six writable shapes from configuration and import-verifies them; reinstating the drop fails exactly one of its cases.

internal/provider/dashboards/dashboard_widgets/interval_resolution.go
```diff
@@ -0,0 +1,224 @@
...
+	return schema.SingleNestedAttribute{
+		Optional: true,
+		Attributes: map[string]schema.Attribute{
+			"auto": schema.SingleNestedAttribute{
+				Optional: true,
+				Attributes: map[string]schema.Attribute{
+					"maximum_data_points": schema.Int64Attribute{
+						Optional:            true,
+						MarkdownDescription: "The most data points to display. The calculated interval keeps within this limit.",
+					},
```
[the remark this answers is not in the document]
dmenilo1: Correct, fixed — both are now bounded to the int32 range via a shared helper, with acceptance cases asserting `2147483648` is rejected on each.

dmenilo1:
@codex review

internal/provider/dashboards/dashboard_widgets/interval_resolution.go
```diff
@@ -0,0 +1,237 @@
...
+
+// IntervalResolutionMaximumDataPointsSchema bounds the value to what the API's
+// int32 field can hold. The conversion is an unchecked cast, so a larger number
+// wraps - 2147483648 arrives as -2147483648 - and the request carries a value
+// the user never wrote.
+func IntervalResolutionMaximumDataPointsSchema(description string) schema.Int64Attribute {
+	return schema.Int64Attribute{
+		Optional: true,
+		Validators: []validator.Int64{
+			int64validator.Between(math.MinInt32, math.MaxInt32),
```
markshleifer-coralogix, line 233: Is it the same as the proto?

markshleifer-coralogix (approved):

internal/provider/dashboards/dashboard_widgets/interval_resolution.go
dmenilo1: No, and on purpose.
The spec says minimum: 1, maximum: 2147483647 — this bound is the full int32 range, so it's wider at the bottom.

The API doesn't enforce the documented minimum: 1, 0, -1, -2147483648, 500 and 2147483647 are all accepted and stored verbatim. 
So matching the proto would reject values the API returns. 
Reading such a dashboard would still work — I checked, validators only run on configuration — but the value couldn't be written back, so an imported dashboard wouldn't be expressible. 
Same call you made for classic decimal in #680.

The bound is only there for what the wire type can't hold: expandInt32Pointer casts unchecked, so 2147483648 wraps to -2147483648 and the request carries a number the user never wrote.

dmenilo1: No, and on purpose. The spec says `minimum: 1, maximum: 2147483647` — this bound is the full int32 range, so it is wider at the bottom.

The API does not enforce the documented minimum: `1`, `0`, `-1`, `-2147483648`, `500` and `2147483647` are all accepted and stored verbatim. So matching the proto would reject values the API returns. Reading such a dashboard would still work — I checked, validators only run on configuration, not on state — but the value could not be written back, so an imported dashboard would not be expressible. Same call you made for classic `decimal` in #680.

The bound is only there for what the wire type cannot hold: `expandInt32Pointer` casts unchecked, so `2147483648` wraps to `-2147483648` and the request carries a number the user never wrote.

Happy to add the documented 1–2147483647 to the attribute description so the guidance still reaches users without the provider enforcing it, the way `DecimalSchema` does. Say the word and I will push it.

PR #691, opened 2026-08-28
fix(dashboard): upgrade state written before schema version 2

markshleifer-coralogix:
## What this fixes

A dashboard whose state is still at schema version 1 cannot be upgraded. The upgrade fails with a value conversion error, so no plan is possible until the resource is removed from state and imported again.

Schema version 1 is what a provider older than v1.13.1 wrote. `upgradeDashboardStateV1ToV2` handed the stored `layout`, `variables`, `filters` and `annotations` to the current schema unchanged, and all four have a different type today. A null value fails the same way, because a stored value keeps the type of the version that wrote it.

Reproduced on release 3.11.0, commit 8b9fee9, with a V1-shaped state:

```
Value Conversion Error … Path: layout
```

So this reaches every released provider that still wires the version 1 upgrader. It is not a regression from a recent change.

## The fix

Version 1 now refreshes the dashboard from the API and lets `flattenDashboard` rebuild every attribute at the current type. That is what versions 2 and 3 already do, so the three upgraders now share one refresh function. Version 1 gets the existing behaviour for a dashboard that no longer exists for free: a warning, a minimal state, and a recreate on the next apply.

The seed each upgrader hands to the refresh carries only the attributes whose type did not change — `id`, `name`, `description`, `folder` and `content_json` — so a folder keeps its id-or-path form and a `content_json` dashboard keeps its configuration.

Versions 2 and 3 also passed the stored `layout`, `variables` and `filters`. `flattenDashboard` replaced all three, so state is unchanged, but a seed that holds no stale type cannot leak one.

## Tests

Every wired prior version is covered, not only the newest, because the version that breaks is the oldest one nobody exercises:

- upgrade from a populated V1 state produces valid V4 state — fails on master
- upgrade from versions 1, 2 and 3 produces valid V4 state — the V1 case fails on master
- a `content_json` dashboard keeps its configuration, for all three versions
- the missing-dashboard warning, extended from versions 2 and 3 to version 1

No schema change, so no documentation change.


markshleifer-coralogix:
## How this was verified end to end

The unit tests build the prior state as `tftypes` values, which skips what production does first: Terraform sends the stored state as JSON and the framework decodes it against the frozen V1 schema. So the fix was also checked through real Terraform, 1.15.6, against a real dashboard.

A version 1 state file was generated from the frozen V1 schema type, so every nested shape is genuine: 24 KB, 73 lists, three levels deep. Its `id` points at a dashboard that exists, and the configuration beside it is the one the mapper produces for that dashboard, so a correct upgrade has to plan empty.

Two runs of `terraform plan` over that same state:

| provider | result |
| --- | --- |
| release 3.11.0 code | `Error: Value Conversion Error … Path: layout` |
| this branch | `No changes. Your infrastructure matches the configuration.` |

An empty plan means the refresh rebuilt `layout`, `variables`, `variables_v2`, `filters`, `annotations`, `auto_refresh`, `time_frame` and `access_policy` correctly at the current type. An earlier run pointed at a deleted dashboard, which covered the other path: the warning, the minimal state, and a plan that offers to recreate it.

The failing attribute in the control run moves between `layout`, `variables_v2` and `filters` from run to run. Framework reflection reports whichever mismatched attribute it reaches first, and map iteration order is random. The bug is every attribute whose type changed, not one of them, which is why the fix refreshes all of them instead of translating one.

cazorla19 (dismissed):

cazorla19 (approved):

PR #699, opened 2026-08-31
feat(dashboards): warn about unknown fields in content_json

dmenilo1:
A dashboard written as `content_json` goes through the SDK decoder, which discards any key it does not recognise. Nothing was reported. A misspelled field looked applied, was thrown away, and the dashboard silently lacked the setting the author thought they had configured.

`content_json` now warns, naming each unrecognised field by its JSON path:

```
Warning: Unknown fields in content_json

  The provider does not recognise these fields and will drop them, so they will
  have no effect on the dashboard:

    layout.sections[0].rows[0].widgets[0].definition.barChart.legend.seriesVisibility
```

## The check mirrors the decoder rather than guessing

A warning on a key the provider actually keeps would be worse than no warning, so the traversal reproduces the decoder's own field matching exactly:

- a key is known when it matches a field's JSON tag **or that tag's `snake_case` spelling** — the decoder accepts both, so reporting `folder_id` alongside `folderId` would be a false alarm
- it descends only into generated model types, the same restriction the decoder applies, so keys inside a free-form map are user data and are left alone
- malformed JSON is left to the existing error path

## Warning, not error — deliberately

An unrecognised key has two possible causes, and only one is the author's mistake. The other is a field newer than the Coralogix SDK this provider is built against. Failing the plan would break configurations that work today and that no configuration change could fix. The message names both causes and says which is likelier for a field copied out of a dashboard export.

## It immediately found a real silent drop

Running the check against every dashboard on a live tenant surfaced `legend.seriesVisibility` on three of them, on both bar and pie charts. That field does not exist anywhere in the pinned SDK, so the provider discards it today: anyone managing one of those dashboards through `content_json` loses the setting with no indication. Closing that gap needs an SDK release and is tracked separately — this PR makes it visible instead of silent, which is the whole point.

Worth stating plainly because it is the practical consequence of merging: users whose dashboards carry such a field will now see a warning on every plan. The warning is accurate, and the alternative is continuing to drop their settings quietly.

## Evidence

- `TestUnknownContentJSONKeys` covers a clean dashboard, a misspelled top-level field, an unknown key nested inside a list element, multiple keys reported in sorted order, both accepted spellings, keys inside an untyped map, and malformed JSON.
- `TestUnknownContentJSONKeysAgreeWithTheDecoder` asserts the stronger property: for content the decoder accepts, nothing is reported. This is the false-positive guard.
- `TestContentJsonValidatorWarnsOnUnknownKeys` covers the wiring — that an unknown key produces exactly one **warning** and no error, that clean content is silent, that malformed JSON still errors, and that the message names the offending key.
- **Negative control**: removing the `snake_case` alias handling makes both the unit test and the agreement test fail, reporting `folder_id` for content the decoder accepts. The tests catch the defect that matters.
- Build, vet, gofmt clean; full unit sweep passes (17 packages); the existing `content_json` and `FromJson` acceptance tests pass unchanged; `make generate` produces no diff, since no schema attribute is added.

## Not included, on purpose

Support for `legend.seriesVisibility`. It needs an SDK release first, and bundling it here would mix a diagnostic with a feature.

Comparing full attribute *paths* rather than leaf names, which would additionally catch a field wired into expand but missing from flatten. That belongs with the systematic parity check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Unreleased
 
 #### resource/coralogix_dashboard
+- FEAT: `content_json` now warns about fields the provider does not recognise, naming each one by its JSON path. Such fields were silently discarded, so a misspelled field looked applied but never reached the API. The check mirrors the decoder's own field matching, including its acceptance of both `camelCase` and `snake_case` spellings, so it does not warn about anything the provider actually keeps. It is a warning rather than an error, because a field can also be newer than the Coralogix SDK the provider is built against, which a configuration change cannot fix.
+
+#### resource/coralogix_dashboard
 - FIX: A metric annotation keeps its `orientation`. The Coralogix UI offers `vertical` and `horizontal` on a metrics-source annotation and the API stores the choice, but the provider had no attribute, so a horizontal annotation read back as vertical and the next apply reset it. Manual and Dataprime annotations already had the attribute. Omit it to accept what the API returns, which is `vertical`.
 
 #### provider
--- a/internal/provider/dashboards/dashboard_schema/common.go
+++ b/internal/provider/dashboards/dashboard_schema/common.go
@@ -17,6 +17,7 @@ package dashboard_schema
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


dmenilo1:
@codex review

markshleifer-coralogix (approved):

PR #703, opened 2026-09-01
feat(dashboards): support widget custom actions

dmenilo1:
A widget can offer actions that open a URL — the Coralogix UI calls them custom actions and puts them in the widget's settings panel, next to units and series management. The API has always stored them. The provider had no attribute, so a dashboard using them could not be written in Terraform, and importing one dropped them silently.

```hcl
widgets = [{
  title = "Avg api response times"
  custom_actions = [{
    name                      = "Open runbook"
    url                       = "https://example.com/runbook"
    data_source               = "logs"
    should_open_in_new_window = true
  }]
  definition = { ... }
}]
```

## The main deviation: nested under the widget, stored on the dashboard

The API keeps actions in a dashboard-level list, each one tagged with the id of the widget it belongs to. Mirroring that shape would have produced bad configuration: a widget's `id` is generated when not set, and Terraform cannot reference a generated value from elsewhere inside the same resource. Users would have had to hand-assign a UUID to every widget they wanted an action on.

So the attribute lives on the widget, which is also where the UI offers it. Expand hoists the actions to the dashboard list and fills `widgetId` from the expanded widget; flatten groups the dashboard list by that id and hands each widget its own back.

The hoist reads widget ids from the **expanded** layout rather than the configuration, because a widget with no configured id gets one generated during expansion, and that generated value is what the request carries. Model and expanded widgets are walked in step, and a length mismatch aborts the hoist rather than attaching an action to the wrong widget.

## Why `data_source` is required

The API requires it, and it rejects the unspecified value — `dataSource: Invalid action data source type` — so it cannot be omitted. The UI never asks for it: it derives the value from the widget's own query.

Deriving it in the provider would mean mapping every widget type's query union to a data source, and the API does not check the value against the widget anyway: writing `metrics` on a logs widget is accepted and stored. Rather than derive a value the API will not validate, the attribute is required and documented as needing to match the widget's query. Deriving it later is a compatible change; guessing wrong now is not.

## Deliberately not included

`goToDashboardAction`, the other arm of the API's action union. The UI's custom action form offers a name and a URL only, so that branch is unreachable through the product, and a URL pointing at another dashboard achieves the same thing. Recorded as api-only in the coverage manifest rather than left unclassified.

`queryId`. The UI does not set it, and no dashboard observed carries it.

An action `id`. The API requires one and does not generate it, so the provider generates a UUID exactly as it already does for a widget without an id. It is not exposed, because it carries no meaning a user would choose.

## Evidence

**Every API rule above came from a probe, not from the generated types.** Each required field was identified by its own error message: `actions[0].id: action id is required`, `actions[0].name: action name not found`, `actions[0].definition: Action definition is empty`, and the data source rejection quoted above. Setting both union arms is refused client-side with `at most one of [customAction, goToDashboardAction] may be set`.

**Verified against an action created in the Coralogix UI**, not only ones this provider wrote. A custom action added through the widget's settings panel stores as a dashboard-level action with `widgetId` matching that widget, and this branch's flatten reads it back as `name="open runbook" url="https://example.com/runbook" data_source="logs" new_window=true`.

`TestAccCoralogixResourceDashboardWidgetCustomActions` covers apply, read, an empty follow-up plan, import, two actions on one widget with a changed field, and removal.

**Two negative controls**, one per direction. Making expand skip the hoist, or making flatten skip the lowering, each fails the first apply. Restoring both returns it to green.

Also green: build, vet, gofmt, the full unit sweep (17 packages), the layout, widget-reference, section-options and annotation-orientation acceptance tests, `content_json` and `FromJson`, REST hydration, and the registry-backed state migration tests. That last one matters because the shared widget attribute map gained a key.

The coverage manifest entry for `ActionDefinition` was stale — it recorded the whole union as unreachable because `Dashboard.actions` was absent from the schema. It now records `customAction` as covered and `goToDashboardAction` as api-only with the reason above.


internal/provider/dashboards/resource_coralogix_dashboard.go
```diff
@@ -4159,8 +4285,62 @@ func flattenDashboardAccessPolicy(planAccessPolicy types.String, accessPolicy *s
...
+}
+
+// flattenWidgetCustomActions returns the actions belonging to one widget. An
+// action the provider cannot express, one whose definition is not a custom
+// action, is left out rather than reported: the dashboard is still readable and
+// only that action is absent.
+func flattenWidgetCustomActions(ctx context.Context, byWidget map[string][]dashboardservice.DashboardAction, widgetID *dashboardservice.UUID) (types.List, diag.Diagnostics) {
+	elemType := types.ObjectType{AttrTypes: widgetCustomActionAttr()}
+	if len(byWidget) == 0 || widgetID == nil || widgetID.Value == nil {
+		return types.ListNull(elemType), nil
```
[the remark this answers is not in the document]
dmenilo1: Correct, thank you — this was a real bug. `custom_actions = []` failed with `Provider produced inconsistent result after apply`, exactly as you describe.

Fixed by rejecting an empty list: `listvalidator.SizeAtLeast(1)`. Omitting the attribute already says the same thing, so an empty list carries no meaning worth accepting.

I tried normalising it to null first, using the `NormalizeEmptyListToNull` plan modifier this repo already has. Terraform refuses that for a non-computed attribute: `planned an invalid value ... planned for absence but config wants existence`. It only works on an `Optional + Computed` attribute, and making this one computed would trade a clear error for "known after apply" on every run.

Added a step asserting the rejection, so the behaviour is pinned either way.

One thing worth passing on: `layout.sections[*].rows[*].widgets[*].definition.data_table.query.metrics.filters[*].operator...label_filters` uses that same plan modifier while being plain `Optional`, so it looks like it has the same latent problem. No empty-list test covers it. Not touching it here, but it is worth a look.

internal/provider/dashboards/dashboard_schema/v4.go
```diff
@@ -113,6 +113,39 @@ func dashboardSchemaAttributesV4() map[string]schema.Attribute {
 														Optional:            true,
 														MarkdownDescription: "Widget title. Required for all inline widgets except markdown, where it is optional.",
 													},
+													// The API stores these on the dashboard, each carrying
+													// the id of the widget it belongs to. They are nested
+													// under the widget here because that is where the
+													// Coralogix UI offers them, and because a widget id is
+													// usually generated, so it cannot be referenced from
+													// elsewhere in the same configuration.
+													"custom_actions": schema.ListNestedAttribute{
```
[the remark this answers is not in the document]
dmenilo1: This is the same claim as on #701, where it was tested and did not hold. There is now a committed regression test for exactly this scenario, and it passes on this branch.

Terraform CLI does call `UpgradeResourceState` even when the stored version matches the current one, and the framework has an explicit passthrough for that case. It unmarshals the raw state against the current type, and `tftypes.jsonUnmarshalObject` ends with:

```go
// make sure we have a value for every attribute
for k, typ := range attrTypes {
    if _, ok := vals[k]; !ok {
        vals[k] = NewValue(typ, nil)
    }
}
```

An attribute in the schema but absent from stored JSON becomes null, so there is nothing to fail on.

`TestAccCoralogixResourceDashboardMigrationFromSameSchemaVersion`, added on #701 and now on master, applies with the released `coralogix/coralogix 3.12.0` and asserts the local provider reads that state as a NoOp with an empty follow-up plan. It passes on this branch, as do the other registry-backed migration tests. Its negative control is that making the read drop an existing attribute fails it with `expected NoOp, got action(s): [update]`.

Introducing V5 would add an upgrader for a type change that does not occur.

The related concern that is real: a *prior* frozen version sharing an attribute helper with the current one would break, because it changes the type of state those versions decode. That is why v1 to v3 each define their widget block inline, and none were touched here.

dmenilo1:
@codex review

internal/provider/dashboards/resource_coralogix_dashboard.go
```diff
@@ -1051,6 +1064,113 @@ func dashboardLogString(dashboard any) string {
...
+		dataSource, ok := dashboardwidgets.ActionDataSourceToProto[m.DataSource.ValueString()]
+		if !ok {
+			diags.AddError("Extract Widget Custom Action Error",
+				fmt.Sprintf("Unknown action data source: %s", m.DataSource.ValueString()))
+			continue
+		}
+		// The API requires an id and does not generate one, so an action that
+		// does not carry one from a previous read gets a fresh id here, exactly
+		// as a widget without an id does.
+		id := uuid.NewString()
```
markshleifer-coralogix: Action IDs change on every apply because expand always generates a new UUID, while flatten does not store the ID. An unrelated dashboard update will therefore replace every action ID. Is it possible to model id as Optional + Computed, preserve it during flatten, and generate one only when it is absent? The dynamic-table link action uses this pattern already.

internal/provider/dashboards/resource_coralogix_dashboard.go
```diff
@@ -1051,6 +1064,113 @@ func dashboardLogString(dashboard any) string {
...
+
+	var model DashboardLayoutModel
+	if dg := layoutModel.As(ctx, &model, basetypes.ObjectAsOptions{}); dg.HasError() {
+		return nil, dg
+	}
+	var sections []SectionModel
+	if dg := model.Sections.ElementsAs(ctx, &sections, true); dg.HasError() {
+		return nil, dg
+	}
+	if len(sections) != len(expanded.Sections) {
```
markshleifer-coralogix, line 1094: Should we return an error here?

dmenilo1:
@codex review

internal/provider/dashboards/resource_coralogix_dashboard.go
dmenilo1: confirmed and fixed.

Worth noting the failure mode this produces, because it's better than an assertion. 
Now that id is in the schema, reverting to per-apply generation fails with Provider produced inconsistent result  after apply — Terraform catches the churn itself, since state holds an id and a regenerated one contradicts the plan. 
So the bug can't come back silently. Added a step that captures the id and asserts it survives an unrelated description change, and confirmed it fails without the fix.

internal/provider/dashboards/resource_coralogix_dashboard.go
dmenilo1: Yes — changed to report, and I did the same at row and widget level, since it's the same invariant.

I traced whether it's reachable: expandDashboardRows and expandDashboardWidgets skip an element on failure but append a diagnostic, and extractDashboard returns on error before the hoist runs. 
So the counts always match today and the branch is defensive.

markshleifer-coralogix (approved):

Issue #689, opened 2026-08-27
Provider: validation and optional allowlisting for the `domain` argument


jmsundar:
### Description

The provider's `domain` argument is an unvalidated free-form string that decides which host receives the API key, and an operator has no supported way to constrain it.

`env` is a closed set. `domain` is not, and the two are not treated consistently.

**1. `domain` has no value validator in the plugin-framework schema.**

In `internal/provider/provider.go` `Schema()`, `env` gets `stringvalidator.OneOfCaseInsensitive(validEnvironmentAliases...)` (line 216). `domain` (lines 221-227) has no equivalent: no host syntax check, no allowlist, nothing rejecting an arbitrary string.

**2. The `domain` validator is self-referential, which looks like a plain typo.**

```go
"domain": schema.StringAttribute{
    Optional: true,
    Validators: []validator.String{
        stringvalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("domain")),
    },
```

Line 224 declares that `domain` conflicts with `domain`. The description on line 226 says "Conflict With 'env'", and the SDKv2 schema above it gets this right (`ConflictsWith: []string{"env"}`, line 119). The framework attribute should reference `"env"`.

**3. Setting `domain` skips the only value check in `Configure()`.**

```go
} else if domain == "" && !slices.Contains(validEnvironmentAliases, terraformEnvironmentAlias) {
    resp.Diagnostics.AddAttributeError(path.Root("env"), "Invalid Coralogix env", ...)
}
```

The enum check (line 326) is guarded by `domain == ""`, so a config that sets `domain` reaches the end of `Configure()` with no value validation at all. From there the value flows straight into the client: `clientset.GrpcTargetFromDomain(domain)` (line 348) and `sdkEnvironment = domain` (line 353), which becomes the OpenAPI/REST client host. `GrpcTargetFromDomain` in `internal/clientset/endpoints.go` normalizes and appends `:443` but does not validate.

**Why this matters for operators**

[...]

PR #714, opened 2026-09-04
CX-56589: add Beta coralogix_fleet_configuration_group resource

cazorla19:
### Description

Add `coralogix_fleet_configuration_group` (resource + data source) for the Fleet Manager Preview configuration-groups API.

The resource manages one nested group: latest family + remote OpenTelemetry Collector YAML. Destroy archives the group (deactivate an active family first). YAML is compared semantically so equivalent documents do not plan. Schema MarkdownDescription includes **Note: This resource is in Beta stage.**

The OpenAPI client is vendored under `internal/openapi/configuration_group_service` until a tagged SDK release ships. Do not `replace` the whole `coralogix-management-sdk` module (that breaks existing resources). Facade sync already landed the generated client on SDK master as `go/openapi/gen/fleet_manager_configuration_groups` ([#689](https://github.com/coralogix/coralogix-management-sdk/pull/689)); this vendor copy can switch to that package on a later SDK bump.

### Acceptance tests

```
$ TF_ACC=1 CORALOGIX_ENV=EU2 go test ./internal/provider/ -run TestAccCoralogixResourceFleetConfigurationGroup -v -timeout 20m

PASS  TestAccCoralogixResourceFleetConfigurationGroup
```

Live EU2 notes (no secrets): create/get/update/import/destroy. API injects `service.version` into `agent_selector`; flatten drops it unless configured. Canonical YAML vs config YAML is ignored on import. Plan is empty after apply.

### Release Note

```release-note
resource/coralogix_fleet_configuration_group: FEAT: Add Beta resource and data source for Fleet Manager configuration groups.
```

### Community Note


cazorla19:
@codex review

cazorla19:
@codex review

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Unreleased
 
+#### resource/coralogix_fleet_configuration_group
+- FEAT: Add Beta `coralogix_fleet_configuration_group` resource and data source for Fleet Manager configuration groups. Destroy archives the group. **Note: This resource is in Beta stage.**
+- DOC: Omitting `family.collector_version` on update keeps the existing remote value; the API has no clear representation.
+- CHORE: Call the Fleet Manager configuration-group API through the management SDK instead of a vendored OpenAPI client.
+- FIX: Omit an unchanged family from replace so group-only updates do not pin stale family IDs.
```
markshleifer-coralogix: I assume that the fixes were temp work. can we have one FEAT line?

internal/clientset/clientset.go
```diff
@@ -227,6 +233,18 @@ func (c *ClientSet) DataEnrichments() (*ess.EnrichmentsServiceAPIService, *cess.
...
 
+func newConfigurationGroupsClient(region, apiKey string) *cfggroups.FleetManagerConfigurationGroupsAPIService {
+	cfg := cfggroups.NewConfiguration()
+	url, found := cxsdkOpenapi.URLFromRegion(strings.ToLower(region))
+	if !found {
+		url = cxsdkOpenapi.URLFromDomain(region)
+	}
+	cfg.Servers = cfggroups.ServerConfigurations{{URL: url}}
+	cfg.AddDefaultHeader("Authorization", "Bearer "+apiKey)
+	cfg.AddDefaultHeader("x-cx-sdk-version", "terraform-"+TF_PROVIDER_VERSION)
```
markshleifer-coralogix, line 244: Is it something all resources have?

internal/provider/fleet/resource_coralogix_fleet_configuration_group.go
```diff
@@ -0,0 +1,793 @@
...
+						Optional:            true,
+						Computed:            true,
+						Default:             booldefault.StaticBool(true),
+						MarkdownDescription: "Whether this family is active. Defaults to true.",
+					},
+					"description": schema.StringAttribute{
+						Optional:            true,
+						MarkdownDescription: "Human-readable family description.",
+					},
+					"collector_version": schema.StringAttribute{
```
markshleifer-coralogix, line 171: `collector_version` is Optional+Computed and has no `UseStateForUnknown`.

If HCL omits this field after apply, the plan value is unknown. Then `familyConfigUnchanged` in `yaml.go` treats it as a family change (`CollectorVersion.Equal`), and `expandReplaceRequest` sends `family` again. This PR says that mints new family/remote IDs.

The example and the acc test always set this field, so that path is untested.

If omit is supported: copy prior state in the plan (or make the field Required). Then add a plan-only step that omits `collector_version` after the first apply.

If omit is not supported: make the field Required and this goes away.

internal/provider/fleet/resource_coralogix_fleet_configuration_group.go
```diff
@@ -0,0 +1,793 @@
...
+	if selector != nil {
+		attrs = selector.Attributes
+	}
+	return flattenStringMap(ctx, selectorAttrsForState(attrs, plan, collectorVersion, dropInjectedSelector), plan)
+}
+
+// The API copies collectorVersion onto agentSelector as service.version when
+// omitted. Drop that injected key on resource reads unless the user configured
+// it, so state matches configuration. Data-source reads keep the remote map.
+func selectorAttrsForState(api map[string]string, plan types.Map, collectorVersion string, dropInjectedSelector bool) map[string]string {
```
markshleifer-coralogix, line 741: Resource flatten drops `agent_selector["service.version"]` when it matches `collector_version` and the user did not set that key. That keeps state in line with HCL when the API injects the key.

Two edge cases:

1. Someone sets that key out of band to the same value as `collector_version`. Refresh hides it. Terraform does not show drift.
2. If the API ever injects a different string, the key stays in state and `agent_selector` can plan forever.

Is dropping the key the intended contract, or should we keep the remote map and document that `service.version` is API-managed?

internal/provider/fleet/resource_coralogix_fleet_configuration_group.go
```diff
@@ -0,0 +1,793 @@
...
+
+func (r *FleetConfigurationGroupResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+	var state *FleetConfigurationGroupResourceModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	id := state.ID.ValueString()
+	if err := deactivateFamilyIfActive(ctx, r.client, state); err != nil {
```
markshleifer-coralogix, line 357: Destroy is two API calls: deactivate the family (`prior` is nil, so replace always sends `family`), then archive.

If deactivate succeeds and archive fails, the group still exists and is inactive. Terraform keeps the resource in state. A later `apply` can read `active = false` and turn it back on, because the schema default is `true`.

A retry of `destroy` should still archive. Acc tests do not assert this path.

Is a failed archive after deactivate an acceptable window, and should Delete say that clearly?

markshleifer-coralogix (commented):
Looks good, a few comments

docs/resources/fleet_configuration_group.md
```diff
@@ -0,0 +1,123 @@
...
+  priority_order = 100
+
+  family = {
+    active            = true
+    collector_version = "0.114.0"
+    description       = "Default production family"
+    metadata = {
+      team = "observability"
+    }
+    remote_configuration = [
```
povilasv, line 43: can you test this with reading from file and giving an example like here?

I think using ` file("")` or other providers do like this:
```
  content = templatefile(
    "config.tftpl",
    {
      collector_id = "prod_collector",
    },
  )
```

docs/resources/fleet_configuration_group.md
```diff
@@ -0,0 +1,123 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "coralogix_fleet_configuration_group Resource - terraform-provider-coralogix"
+subcategory: ""
+description: |-
+  Fleet Manager configuration group with its latest family and remote OpenTelemetry Collector YAML. Destroy archives the group. Note: This resource is in Beta stage.
+---
+
+# coralogix_fleet_configuration_group (Resource)
+
+Fleet Manager configuration group with its latest family and remote OpenTelemetry Collector YAML. Destroy archives the group. **Note: This resource is in Beta stage.**
```
povilasv: Let's also mention private preview

examples/resources/coralogix_fleet_configuration_group/resource.tf
```diff
@@ -0,0 +1,53 @@
...
+  family = {
+    active            = true
+    collector_version = "0.114.0"
+    description       = "Default production family"
+    metadata = {
+      team = "observability"
+    }
+    remote_configuration = [
+      {
+        name              = "default"
```
povilasv:
```suggestion
        name              = "otel-agent"
```

maybe we should give example with multiple collectors?

One could target agent, another cluster-collector or smth like this

docs/resources/fleet_configuration_group.md
cazorla19: Done in d129ac0. The example now loads collector YAML with `file("./otel-agent.yaml")` and `file("./otel-cluster-collector.yaml")`, matching the recording-rules pattern in this repo. `TestAccCoralogixResourceFleetConfigurationGroupFromFile` applies that same `file()` config (agent + cluster-collector remotes, empty follow-up plan, import). `TestExampleCollectorYAMLFilesParse` also checks the example YAML files parse.

docs/resources/fleet_configuration_group.md
cazorla19: Done in d129ac0. Schema MarkdownDescription now says this resource is in private preview (Beta), and the generated resource/data-source docs pick that up.

examples/resources/coralogix_fleet_configuration_group/resource.tf
cazorla19: Follow-up in d129ac0: the example now has two remotes, otel-agent (cx.agent.type = agent) and otel-cluster-collector (cx.agent.type = cluster-collector).

CHANGELOG.md
cazorla19: Collapsed to one FEAT line in d195e73. Left the example DOC from the file-based YAML follow-up.

internal/clientset/clientset.go
cazorla19: gRPC resources send it via terraformSDKCallPropertiesCreator, and OpenAPI resources that go through cs.* get it from WithTerraformVersion. This client is built outside cs.* because the pinned SDK ClientSet still has no ConfigurationGroups(), so the same terraform-<version> header is set here. SCIM users / Grafana do not.

internal/provider/fleet/resource_coralogix_fleet_configuration_group.go
cazorla19: Omit is supported: the API keeps the prior value, and empty string is not a clear. d195e73 copies prior state with UseStateForUnknown, treats an unknown plan value as unchanged so group-only updates do not resend family, and adds a plan-only acc step that omits collector_version after the first apply.

internal/provider/fleet/resource_coralogix_fleet_configuration_group.go
cazorla19: Dropping the injected key is the intended contract, so HCL that does not set service.version stays without it in resource state. (1) Same-value out-of-band writes are indistinguishable from injection, so refresh hides them. (2) A different injected string is kept and will plan, which is the signal that the API wrote something other than collector_version. Data-source reads keep the remote map. Documented on agent_selector in d195e73.

internal/provider/fleet/resource_coralogix_fleet_configuration_group.go
cazorla19: Yes, that window is acceptable. Archive rejects an active family, so delete is deactivate then archive. If archive fails, the group stays inactive in Coralogix and in Terraform state; retrying destroy still archives (deactivate no-ops when already inactive). A later apply is supposed to keep managing the resource, so it will send active=true from the schema default. Documented on the resource and in Delete in d195e73. No acc coverage for an injected archive failure.

markshleifer-coralogix (approved):

